PR #27, opened 2021-03-23
CLI: Code Revision - Use omegaconf to replace hydra for configuration

abuccts:
**Major Revisions**
* remove hydra
* use omegaconf to merge configurations

### DIFF
--- a/superbench/benchmarks/__init__.py
+++ b/superbench/benchmarks/__init__.py
@@ -3,9 +3,10 @@
 
 """Exposes interfaces of benchmarks used by SuperBench executor."""
 
-from .return_code import ReturnCode
-from .context import Platform, Framework, Precision, ModelAction, BenchmarkType, BenchmarkContext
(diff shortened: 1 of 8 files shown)


omry:
Howdy,
curious about the motivation here.
For all but the most primitive use cases, I definitely recommend Hydra over OmegaConf.

abuccts:
Hi @omry, sorry for the late response.

The key feature we need from Hydra is overriding config files through a parameter string, e.g., user can use `--config-file` and `--config-override` to specify config file and override it on the fly, so OmegaConf may be enough for this case.

The main concern is we need a flexible command line interface for this tool, and we only want Hydra for several (not all) commands and only a part of configuration (some config like password should not be written to config files).

omry:
Take a look at the Hydra [Compose API](https://hydra.cc/docs/experimental/compose_api).

It provides access to the Hydra config composition capability without integrating with the command line.

PR #32, opened 2021-04-07
Benchmarks: Code Revision - Revise result process interface and add result checking

guoshzhao:
Major Revisions
1. Modify result process interface to add more check and return True/False.
2. For MicroBenchmark, there have two function, _process_numeric_result() is to process result from python scripts. _process_raw_result is to process raw string result from system call benchmarks.
3. For DockerBenchmark, only _process_raw_result is needed to process raw string result.
4. For ModelBenchmark, only __process_model_result is needed to process result from python scripts.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/micro_base.py
+++ b/superbench/benchmarks/micro_benchmarks/micro_base.py
@@ -5,6 +5,7 @@
 
 from abc import abstractmethod
 
+from superbench.common.utils import logger
 from superbench.benchmarks import BenchmarkType
 from superbench.benchmarks.base import Benchmark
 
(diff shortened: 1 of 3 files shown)


cp5555:
Add description following:

**Major Revisions**
- [Revision 1]
- [Revision 2]

**Minor Revisions**
- [Revision 1]
- [Revision 2]

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -40,17 +41,46 @@ def _preprocess(self):
...
+
+    def _process_raw_result(self, raw_output):
         """Function to process raw results and save the summarized results.
 
         Args:
-            output (str): raw output string of the micro-benchmark.
+            raw_output (str): raw output string of the micro-benchmark.
+
+        Return:
+            True if the raw output string is valid and result can be extracted.
```
abuccts, line 80: will this function save results? how does it know where to save?

abuccts (approved):

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: This function also need to call the following function to save result:
        self._result.add_raw_data()
        self._result.add_result()
       
Benchmark owner need to know this. I will add more comments.

guoshzhao:
> Add description following:
> 
> **Major Revisions**
> 
> * [Revision 1]
> * [Revision 2]
> 
> **Minor Revisions**
> 
> * [Revision 1]
> * [Revision 2]

Done.

PR #34, opened 2021-04-08
Executor: Init - Add superbench executor class

abuccts:
__Major Revisions__
* add executor class
* update default config to exec benchmarks

### DIFF
--- a/superbench/config/default.yaml
+++ b/superbench/config/default.yaml
@@ -1,25 +1,22 @@
-# @package _global_
-
-# Hydra config
-hydra:
-  run:
-    dir: ./outputs/${now:%Y-%m-%d_%H-%M-%S}
-  sweep:
-    dir: ./outputs/${now:%Y-%m-%d_%H-%M-%S}
-  job_logging:
-    formatters:
-      colorlog:
-        format: >-
-          [%(cyan)s%(asctime)s %(hostname)s%(reset)s][%(blue)s%(filename)s:%(lineno)s%(reset)s][%(log_color)s%(levelname)s%(reset)s] %(message)s
-defaults:
-  - hydra/job_logging: colorlog
-  - hydra/hydra_logging: colorlog
-
-# SuperBench config
+# SuperBench Config
 superbench:
-  use: []
+  use: null
   benchmarks:
-    pytorch_models:
-      enable: true
+    bert_models:
+      use: on
+      frameworks:
+        - pytorch
+      models:
+        - bert-base
+        - bert-large
       parameters:
+        duration: 0
+        num_warmup: 64
+        num_steps: 2048
         batch_size: 32
+        precision:
+          - float32
+          - float16
+        model_action:
+          - train
+          - inference
--- a/superbench/config/default.yaml
+++ b/superbench/config/default.yaml
@@ -1,9 +1,9 @@
 # SuperBench Config
 superbench:
-  use: null
+  enable: null
   benchmarks:
     bert_models:
-      use: on
+      enable: true
       frameworks:
         - pytorch
       models:
--- a/superbench/config/default.yaml
+++ b/superbench/config/default.yaml
@@ -2,6 +2,18 @@
 superbench:
   enable: null
   benchmarks:
+    matmul:
+      enable: true
+      frameworks:
+        - pytorch
+      parameters:
+        num_steps: 2048
+    sharding_matmul:
+      enable: true
+      frameworks:
+        - pytorch
+      parameters:
+        num_steps: 2048
     bert_models:
       enable: true
       frameworks:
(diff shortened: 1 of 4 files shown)


superbench/config/default.yaml
```diff
@@ -1,25 +1,22 @@
...
-# SuperBench config
+# SuperBench Config
 superbench:
-  use: []
+  use: null
   benchmarks:
-    pytorch_models:
-      enable: true
+    bert_models:
+      use: on
```
cp5555, line 6: use or enable?

superbench/config/default.yaml
```diff
@@ -1,25 +1,22 @@
...
+      frameworks:
+        - pytorch
+      models:
+        - bert-base
+        - bert-large
```
cp5555, line 11: I remember we have a discussion before. One test for one model.

superbench/executor/executor.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        self._sb_config = sb_config
+        self._docker_config = docker_config
+        self._output_dir = output_dir
+
+        self.__set_logger('sb-exec.log')
+        logger.info('Executor uses config: %s.', self._sb_config)
+        logger.info('Executor writes to: %s.', self._output_dir)
+
+        self.__validate_sb_config()
```
cp5555, line 32: `self.__validate_sb_config()` should return true if it pass the validation. If not, what should we do?

superbench/executor/executor.py
```diff
@@ -0,0 +1,113 @@
...
+        Return:
+            list: List of benchmarks which will be executed.
+        """
+        if self._sb_config.superbench.use:
+            if isinstance(self._sb_config.superbench.use, str):
+                return [self._sb_config.superbench.use]
+            elif isinstance(self._sb_config.superbench.use, (list, ListConfig)):
+                return list(self._sb_config.superbench.use)
+        # TODO: may exist order issue
+        return [k for k, v in self._sb_benchmarks.items() if v.use]
```
cp5555, line 61: Do we need to check whether `_sb_benchmarks` is a dict. If it doesn't match anything, it can return None.

superbench/executor/executor.py
```diff
@@ -0,0 +1,113 @@
...
+            elif isinstance(val, (list, ListConfig)):
+                argv.append('--{} {}'.format(name, ' '.join(val)))
+            elif isinstance(val, bool) and val:
+                argv.append('--{}'.format(name))
+        return ' '.join(argv)
+
+    def exec(self):
+        """Run the SuperBench benchmarks locally."""
+        for benchmark_name in self._sb_benchmarks:
+            if benchmark_name not in self._sb_enabled:
```
cp5555, line 87: Why do we need to keep self._sb_enabled? Can we check it directly?

superbench/executor/executor.py
```diff
@@ -0,0 +1,113 @@
...
+                continue
+            benchmark_config = self._sb_benchmarks[benchmark_name]
+            if benchmark_name.endswith('_models'):
+                for framework in benchmark_config.frameworks:
+                    for model in benchmark_config.models:
+                        logger.info('Executor is going to execute %s: %s/%s.', benchmark_name, framework, model)
+                        context = BenchmarkRegistry.create_benchmark_context(
+                            model,
+                            platform=Platform.CUDA,
```
cp5555, line 96: Can we input the platform?

guoshzhao: add TODO.

superbench/config/default.yaml
guoshzhao: enable

superbench/executor/executor.py
```diff
@@ -0,0 +1,113 @@
...
+            if isinstance(val, (str, int, float)):
+                argv.append('--{} {}'.format(name, val))
+            elif isinstance(val, (list, ListConfig)):
+                argv.append('--{} {}'.format(name, ' '.join(val)))
+            elif isinstance(val, bool) and val:
+                argv.append('--{}'.format(name))
+        return ' '.join(argv)
+
+    def exec(self):
+        """Run the SuperBench benchmarks locally."""
```
guoshzhao, line 85: Here the exec will be called by sb runner when do distributed testing?

guoshzhao (approved):

superbench/executor/executor.py
abuccts: updated comment

superbench/executor/executor.py
abuccts: will be checked in `__validate_sb_config`

superbench/executor/executor.py
abuccts: yes

cp5555 (approved):

PR #36, opened 2021-04-08
Benchmarks: Add Benchmark - Add op-sharding microbenchmark, including matmul and sharding_matmul.

guoshzhao:
Major Features:
- Realize ShardingMatmul class.
- Add examples for matmul and sharding_matmul.

Minor Features:
- Add commands into comments for examples.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/sharding_matmul.py
+++ b/superbench/benchmarks/micro_benchmarks/sharding_matmul.py
@@ -1,0 +1,239 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Module of the ShardingMatmul benchmarks."""
+
+import os
+import time
+
+import torch
+
+from superbench.common.utils import logger
+from superbench.benchmarks import BenchmarkRegistry, ReturnCode
+from superbench.benchmarks.micro_benchmarks import MicroBenchmark
+from superbench.benchmarks.context import Enum
+
+
+class ShardingMode(Enum):
+    """The Enum class representing different sharding mode."""
+    ALLREDUCE = 'allreduce'
+    ALLGATHER = 'allgather'
+    NOSHARDING = 'nosharding'
+
+
+class ShardingMatmul(MicroBenchmark):
+    """The base class of micro-benchmarks."""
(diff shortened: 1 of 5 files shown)


superbench/benchmarks/micro_benchmarks/sharding_matmul.py
```diff
@@ -0,0 +1,239 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Module of the ShardingMatmul benchmarks."""
```
cp5555, line 4: We would write a detailed description here about what the purpose for this microbenchmark.

superbench/benchmarks/micro_benchmarks/sharding_matmul.py
```diff
@@ -0,0 +1,239 @@
...
+
+import os
+import time
+
+import torch
```
cp5555, line 9: One question: If our microbenchmark depends on PyTorch. Do we need to check it? Because we don't run PyTorch if we don't enable it but we don't check in microbenchmark.

superbench/benchmarks/micro_benchmarks/sharding_matmul.py
```diff
@@ -0,0 +1,239 @@
...
+        torch.backends.cudnn.benchmark = True
+
+    def add_parser_arguments(self):
+        """Add the specified arguments."""
+        super().add_parser_arguments()
+
+        self._parser.add_argument(
+            '--n',
+            type=int,
+            default=12288,
```
cp5555, line 47: Shall we change this setting?

abuccts (approved):

superbench/benchmarks/micro_benchmarks/sharding_matmul.py
guoshzhao: Done.

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/sharding_matmul.py
guoshzhao: Have added TODO.

PR #37, opened 2021-04-09
Benchmarks: Fix Bug - Specify the framework 'pytorch' for matmul example.

guoshzhao:


examples/benchmarks/matmul.py
```diff
@@ -3,17 +3,16 @@
 
 """Model benchmark example for matmul with pytorch.
 
-Command to run:
+Commands to run:
```
abuccts, line 6: why "Commands"?

abuccts (approved):

examples/benchmarks/matmul.py
guoshzhao: Just to keep the same with other examples, like bert-large.

cp5555 (approved):

PR #38, opened 2021-04-09
Runner: Init - Add superbench runner class

abuccts:
__Major Revisions__
* init runner class


tests/runner/test_runner.py
```diff
@@ -0,0 +1,42 @@
...
+    def tearDown(self):
+        """Hook method for deconstructing the test fixture after testing it."""
+        shutil.rmtree(self.output_dir)
+
+    def test_set_logger(self):
+        """Test log file exists."""
+        expected_log_file = Path(self.runner._output_dir) / 'sb-run.log'
+        self.assertTrue(expected_log_file.is_file())
+
+    def test_run(self):
```
cp5555, line 39: It's a little bit strange to check NotImplementedError

cp5555 (approved):

tests/runner/test_runner.py
abuccts: removed

tests/runner/test_runner.py
```diff
@@ -0,0 +1,41 @@
...
+from omegaconf import OmegaConf
+
+from superbench.runner import SuperBenchRunner
+
+
+class RunnerTestCase(unittest.TestCase):
+    """A class for runner test cases.
+
+    Args:
+        unittest.TestCase (unittest.TestCase): TestCase class.
```
guoshzhao, line 20: This is inheriting not args, right?

guoshzhao (approved):

tests/runner/test_runner.py
abuccts: updated

PR #39, opened 2021-04-09
Benchmarks: Add Benchmark - Add computation and communication overlap micro benchmark

yukirora:
Major Features:
-  Add microbenchmark for computation and communication overlap.
-  Add related test for 1GPU case and multi GPUs case.


superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,234 @@
...
+            default=30,
+            required=False,
+            help="The execution ratio number between computation kernel and nccl kernel",
+        )
+        self._parser.add_argument(
+            '--kernel',
+            type=ComputationKernelType,
+            default=[ComputationKernelType.MUL],
+            required=False,
+            help='Computation kernel type. E.g. {}.'.format(' '.join(ComputationKernelType.get_values())),
```
abuccts, line 97: accept one kernel or a list of kernels?

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,234 @@
...
+            '--num_warmup',
+            type=int,
+            default=10,
+            required=False,
+            help='The number of warmup step.',
+        )
+        self._parser.add_argument(
+            '--num_steps',
+            type=int,
+            default=2000,
```
abuccts, line 109: use 64 warmup and 2048 steps?

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,234 @@
...
+        torch.distributed.destroy_process_group()
+        return True
+
+
+BenchmarkRegistry.register_benchmark(
+    'pytorch-computation-communication-overlap-matmul', ComputationCommunicationOverlap, parameters='--kernel matmul'
+)
+BenchmarkRegistry.register_benchmark(
+    'pytorch-computation-communication-overlap-mul', ComputationCommunicationOverlap, parameters='--kernel mul'
+)
```
abuccts, line 234: @guoshzhao can we merge them into one benchmark here?

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def __kernel_nccl_pipeline(self, kernel, matA, matB, stages, message, times):
+        """computation and nccl kernel pipeline with single GPU.
+
+        Args:
+            kernel (ComputationKernelType): the type of the computation kernel to run
+            matA (list[tensor]): the matrix list used in matmul or mul for every stage
+            matB (tensor): the matrix used in matmul
+            stages (int): the ratio number of computation kernel and communication kernel
+            message(tensor): the data used to be transfered through NCCL
```
abuccts, line 145:
```suggestion
            message(tensor): the data used to be transferred through NCCL
```

abuccts (commented):
pls run the following to fix CI issues
```sh
# install
python3 -m pip install -e .[dev,test]

# format code
python3 setup.py format

# check code style
python3 setup.py lint

# run all unit tests
python3 setup.py test
```

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,234 @@
+# Copyright (c) Microsoft Corporation.
```
cp5555, line 1: Would you please add test? Otherwise, it cannot pass the check.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+            matB (tensor): the matrix used in matmul
+            stages (int): the ratio number of computation kernel and communication kernel
+            message(tensor): the data used to be transferred through NCCL
+            times(int): number of times in one step to run
+
+        Return:
+            True of False: if computation kernel type is invalid, return False, else, return True
+        """
+        for i in range(times):
+            if self.__world_size > 1:
```
cp5555, line 157: If world_size ==1, shall we log error and return false? Maybe we should check this in the _benchmark and we don't need to do the check each time in for loop

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+                torch.distributed.all_reduce(message, op=torch.distributed.ReduceOp.SUM, async_op=True)
+            for stage in range(stages):
+                if kernel == ComputationKernelType.MUL:
```
cp5555, line 160: Do we need to do the check in the for loop?

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+        # warm up
+        for i in range(self._args.num_warmup):
+            if not self.__kernel_nccl_pipeline(kernel, MatA, MatB, self._args.ratio, message, times=100):
+                return False
+        torch.cuda.synchronize()
+
+        # run and collect results
+        start = time.perf_counter()
+        for i in range(self._args.num_steps):
+            self.__kernel_nccl_pipeline(kernel, MatA, MatB, self._args.ratio, message, times=100)
```
cp5555, line 209: Really strange because NCCL is also launched in this function, but we would like to know the computation kernel time in this function. Why do we measure the completion of `self.__kernel_nccl_pipeline()`?

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+            stages (int): the ratio number of computation kernel and communication kernel
+            message(tensor): the data used to be transferred through NCCL
+            times(int): number of times in one step to run
+
+        Return:
+            True of False: if computation kernel type is invalid, return False, else, return True
+        """
+        for i in range(times):
+            if self.__world_size > 1:
+                torch.distributed.all_reduce(message, op=torch.distributed.ReduceOp.SUM, async_op=True)
```
cp5555, line 158: The ratio between NCCL and Computation is very critical for this testing. For example, in K80, the computation will be very slow. Do we need to keep different setting for different GPU? @guoshzhao .

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
yukirora: I mean to measure the total duration from computation start to computation end，when the self.__kernel_nccl_pipeline() complete, it means all the computation have completed because they are not async. I think we can't get computation kernel time directly if we don't use profiling tools like nsight.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
guoshzhao: Yes, I think it's better to merge them into one benchmark. Just like allredue and allgather in ShardingMatmul. Just register it as:
BenchmarkRegistry.register_benchmark(
    'pytorch-computation-communication-overlap', ComputationCommunicationOverlap, parameters='--kernel matmul mul'
)

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+    """The base class of micro-benchmarks."""
+    def __init__(self, name, parameters=''):
+        """Constructor.
+
+        Args:
+            name (str): benchmark name.
+            parameters (str): benchmark parameters.
+        """
+        super().__init__(name, parameters)
+        # Command lines to launch the micro-benchmarks.
```
guoshzhao, line 45: Please remove the member self.__commands, I will remove it in 'ShardingMatmul' too.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+            default=1880,
+            required=False,
+            help='The N dim of matmul (M, K) * (K, N).',
+        )
+        self._parser.add_argument(
+            '--k',
+            type=int,
+            default=1024,
+            required=False,
+            help='The K dim of matmul (M, K) * (K, N) or The K dim or mul (M, K)*(M, K). ',
```
guoshzhao, line 67: the second 'or' should be 'of', and there has one redundant space char in the end.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+        self._parser.add_argument(
+            '--m',
+            type=int,
+            default=640,
+            required=False,
+            help='The M dim of matmul (M, K) * (K, N)  or The m dim or mul (M, K)*(M, K). ',
```
guoshzhao, line 74: same with last comment.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+        self._parser.add_argument(
+            '--p',
+            type=int,
+            default=5000,
+            required=False,
+            help='The P dim of Tensor(P, Q) to transfer in Nccl AllReduce.',
```
guoshzhao, line 81: Use the persistent word for 'NCCL'? I saw 'Nccl', 'NCCL' and 'nccl' in the file. Please unify them.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
guoshzhao: To merge matmul and mat as one benchmark, here should be list of kernels.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
yukirora: my concern is should we always use the same shape for all kernels?

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+        if torch.cuda.is_available():
+            torch.cuda.set_device(self.__local_rank)
+
+        return True
+
+    def __kernel_nccl_pipeline(self, kernel, matA, matB, stages, message, times):
+        """Computation and nccl kernel pipeline with single GPU.
+
+        Args:
+            kernel (ComputationKernelType): the type of the computation kernel to run
```
guoshzhao, line 146: Pls add dot '.' at the end of line.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
guoshzhao: ratio is a argument, I think it's the right way to control the benchmark behavior for different GPUs.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
guoshzhao: Agree. If self.__world_size <= 1 is meaningless for this benchmark, then we should log error to tell user the settings are wrong.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
yukirora: my concern is should we always use the same shape for all kernels? @guoshzhao

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+
+        compute_metric = 'computation_communication_overlap_gpu{}_computation_cost_{}'.format(self.__local_rank, kernel)
+        total_metric = 'computation_communication_overlap_gpu{}_total_cost_{}'.format(self.__local_rank, kernel)
+        compute_elapse_times = [(compute_end - start) * 1000 / self._args.num_steps]
+        elapse_times = [(end - start) * 1000 / self._args.num_steps]
+        if not self._process_numeric_result(compute_metric, compute_elapse_times
+                                            ) or not self._process_numeric_result(total_metric, elapse_times):
+            return False
+
+        logger.info(
```
guoshzhao, line 228: It's better to log benchmark name (self._name) and the time of run (self._curr_run_index) here. We can config how many times to run the benchmark, self._curr_run_index is the current index of running.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,242 @@
...
+        torch.cuda.synchronize()
+        end = time.perf_counter()
+
+        logger.info(
+            'GPU {} total compute cost {}'.format(
+                self.__local_rank, (compute_end - start) * 1000 / self._args.num_steps
+            )
+        )
+
+        compute_metric = 'computation_communication_overlap_gpu{}_computation_cost_{}'.format(self.__local_rank, kernel)
```
guoshzhao, line 220: Why do we need two similiar metrics? Which metric will show the unusual results? or both can?

yukirora: compute_metric is the metric for computation end, if any gpu has issue, it will show unusual for different gpu. Another metric is for both computation and communication end, it is almost same for all gpus in a machine at any time, it's a more stable value, so in my opinion it's used to check if this machine has performance downgrade for different machines.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
guoshzhao: If we want to use another shape, we can add a same benchmark with different settings in user configs. It means your benchmark can be run two times with different shapes.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
guoshzhao: Thanks! Let's have a discussion on monday.

guoshzhao:
Major Features:

Minor Features:

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,255 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Module of the ComputationCommunicationOverlap benchmarks.
+
+ComputationCommunicationOverlap benchmark is used to test the performance of single node
+when communication and computation overlap and figure out the issued GPU.
+if the node has performance downgrade on this context:
```
guoshzhao, line 8: Uppercase for first char.

tests/benchmarks/micro_benchmark/test_computation_communication_overlap.py
```diff
@@ -0,0 +1,166 @@
...
+        if torch.cuda.is_available():
+            return func(*args, **kwargs)
+
+        message = 'No cuda device'
+        logger.error('Device error -  message: {}.'.format(message))
+
+    return wrapper
+
+
+@skip_if_not_multigpu
```
guoshzhao, line 104: @abuccts to see whether we should have unified decorators.

guoshzhao (approved):

tests/benchmarks/utils.py
```diff
@@ -4,6 +4,11 @@
 """Utilities for benchmark tests."""
```
cp5555, line 4: Shall we separate the PR for utils?

tests/benchmarks/micro_benchmarks/test_computation_communication_overlap.py
```diff
@@ -0,0 +1,92 @@
...
+def skip_if_not_multigpu(func):
+    """Multi-GPU tests requires at least 2 GPUS. Skip if this is not met."""
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        if torch.cuda.is_available() and torch.cuda.device_count() >= 2:
+            return func(*args, **kwargs)
+        message = 'Need at least {} CUDA devices'.format(2)
+        logger.error('Device error -  message: {}.'.format(message))
+
+    return wrapper
```
cp5555, line 25: @abuccts what's your thought for this?

tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
```
cp5555, line 1: Separate the PR?

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,255 @@
...
+        Args:
+            name (str): benchmark name.
+            parameters (str): benchmark parameters.
+        """
+        super().__init__(name, parameters)
+        self.__world_size = 1
+        self.__local_rank = 0
+        torch.backends.cudnn.benchmark = True
+
+    def add_parser_arguments(self):
```
cp5555, line 49: I remember we have some reference for the default setting of matmul and mul. Shall we add the comment for reference?

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -0,0 +1,255 @@
...
+                ratio between computation kernel and NCCL kernel={}'.format(
+                    self.__world_size, M, K, N, P, Q, self._args.ratio
+                )
+            )
+
+        MatA = list()
+        MatB = list()
+        # Matrix A
+        for _ in range(self._args.ratio):
+            MatA.append(torch.randn(M, K).cuda())
```
cp5555, line 195: Do we handle OOM issue if M,K,N are too large?

tests/benchmarks/utils.py
cp5555, line 4: Shall we separate the PR?

tests/benchmarks/utils.py
```diff
@@ -22,3 +27,49 @@ def clean_simulated_ddp_distributed_env():
...
+
+def get_free_port():
+    """Get free port."""
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        s.bind(('', 0))
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        return s.getsockname()[1]
+
+
+def setup_simulated_ddp_distributed_env_custom(world_size, local_rank, port):
```
guoshzhao, line 40: Please modify my tests to merge setup_simulated_ddp_distributed_env function into one.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
yukirora: It will be handled in executor outside

tests/benchmarks/utils.py
yukirora: sure

tests/benchmarks/utils.py
yukirora: I will modify it in another pr for revise test util

tests/__init__.py
yukirora: sure

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
yukirora: The shape is what I randomly try and do the test on A100, later we may provide the suggested default configuration for different device.

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
yukirora: delete this condition after dicussion

cp5555 (approved):

PR #40, opened 2021-04-11
Benchmarks: Add Feature - Add _post_process() interface.

guoshzhao:
Major changes:
- add _post_process() interface in base.py.
- define new return code: DISTRIBUTED_SETTING_DESTROY_FAILURE.


superbench/benchmarks/micro_benchmarks/sharding_matmul.py
```diff
@@ -221,7 +217,7 @@ def _benchmark(self):
         K = self._args.k
         N = self._args.n
         for mode in self._args.mode:
-            if mode == ShardingMode.NOSHARDING or self.__world_size == 1:
+            if mode == ShardingMode.NOSHARDING:
```
cp5555, line 220: Why remove self.__world_size == 1?

superbench/benchmarks/return_code.py
```diff
@@ -18,7 +18,8 @@ class ReturnCode(Enum):
     MODEL_TRAIN_FAILURE = 11
     MODEL_INFERENCE_FAILURE = 12
     DISTRIBUTED_SETTING_INIT_FAILURE = 13
-    DATASET_GENERATION_FAILURE = 14
-    DATALOADER_INIT_FAILURE = 15
-    OPTIMIZER_CREATION_FAILURE = 16
-    MODEL_CREATION_FAILURE = 17
+    DISTRIBUTED_SETTING_DESTROY_FAILURE = 14
```
cp5555, line 21: Why not add it as 18 directly?

superbench/benchmarks/micro_benchmarks/sharding_matmul.py
```diff
@@ -14,7 +14,7 @@
 import os
```
cp5555, line 14: Shall we separate it in another PR? It's not the interface actually.

superbench/benchmarks/micro_benchmarks/sharding_matmul.py
guoshzhao: The reason is introduced in the description of PR:

> For self.__matmul_nosharding() change the condition from 'if mode == ShardingMode.NOSHARDING or self.__world_size == 1:' to 'if mode == ShardingMode.NOSHARDING', since sharding-matmul with 1GPU should also be OK to run.

superbench/benchmarks/return_code.py
guoshzhao: Just put it near the DISTRIBUTED_SETTING_INIT_FAILURE error.

superbench/benchmarks/micro_benchmarks/sharding_matmul.py
guoshzhao: Done.

cp5555 (approved):

abuccts (approved):

PR #51, opened 2021-04-12
Executor - Fix issues when executing benchmarks

abuccts:
__Major Revisions__

* fix missing package in dockerfile
* update benchmark list and parameters
* catch runtime errors
* refine logging info

### DIFF
--- a/superbench/config/default.yaml
+++ b/superbench/config/default.yaml
@@ -7,7 +7,7 @@ superbench:
       frameworks:
         - pytorch
       parameters:
-        num_steps: 256
+        num_steps: 2048
     bert_models:
       enable: true
       frameworks:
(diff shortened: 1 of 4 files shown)


superbench/config/default.yaml
```diff
@@ -7,13 +7,7 @@ superbench:
...
         - pytorch
       parameters:
-        num_steps: 2048
-    sharding_matmul:
-      enable: true
-      frameworks:
-        - pytorch
-      parameters:
-        num_steps: 2048
+        num_steps: 256
```
cp5555, line 10: Why do we remove sharing_matmul in default?

superbench/config/default.yaml
```diff
@@ -23,9 +17,9 @@ superbench:
         - bert-large
       parameters:
         duration: 0
-        num_warmup: 64
-        num_steps: 2048
-        batch_size: 32
+        num_warmup: 8
+        num_steps: 256
+        batch_size: 16
```
cp5555, line 22: Do we need to reduce all of them?

cp5555 (approved):

guoshzhao (approved):

PR #58, opened 2021-04-14
Benchmarks: Code Revision - Fix some issue for BERT benchmark.

guoshzhao:
**Major Revision**
- Add self._gpu_available judgement in inference process.
- Set seq_length as 128 for BERT example.

### DIFF
--- a/examples/benchmarks/pytorch_bert_large.py
+++ b/examples/benchmarks/pytorch_bert_large.py
@@ -15,7 +15,7 @@ if __name__ == '__main__':
     # Create context for bert-large benchmark and run it for 120 * 2 seconds.
     context = BenchmarkRegistry.create_benchmark_context(
         'bert-large',
-        parameters='--batch_size 1 --duration 120 --seq_len 8 --precision float32 --run_count 2',
+        parameters='--batch_size 1 --duration 120 --seq_len 128 --precision float32 --run_count 2',
         framework=Framework.PYTORCH
     )
 
--- a/examples/benchmarks/pytorch_bert_large.py
+++ b/examples/benchmarks/pytorch_bert_large.py
@@ -1,22 +1,35 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-"""Model benchmark example for bert-large.
+"""Model benchmark example for bert-large (24-layer, 1024-hidden, 16-heads, 340M parameters).
 
 Commands to run:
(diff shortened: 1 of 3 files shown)


cp5555 (approved):

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -171,11 +171,12 @@ def _inference_step(self, precision):
             self._model.eval()
             while True:
                 for idx, sample in enumerate(self._dataloader):
-                    torch.cuda.synchronize()
                     start = time.time()
-                    sample = sample.cuda()
+                    if self._gpu_available:
+                        sample = sample.cuda()
```
abuccts, line 176: add tests with `--no-gpu`?

guoshzhao: Better to choose one cheap model to test with '--no-gpu', I will use CNN model to test this option.

abuccts: use `--batch-size 0` and `--no-gpu`? so that it will be covered in tests but won't run actually

guoshzhao: I have add --no-gpu test case in CNN tests, please check.

examples/benchmarks/pytorch_bert_large.py
```diff
@@ -1,22 +1,35 @@
...
+
+from superbench.benchmarks import Platform, Framework, BenchmarkRegistry
 from superbench.common.utils import logger
 
 if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--distributed', action='store_true', default=False, help='Whether to enable distributed training.'
+    )
+    args = parser.parse_args()
```
abuccts, line 22: expose all parameters to args? you could have `parameters=' '.join(sys.argv[1:])`
the example will be run as
```
python3 -m torch.distributed.launch --nproc_per_node=8 examples/benchmarks/pytorch_bert_large.py \
  --batch_size 1 --duration 120 --seq_len 128 --precision float32 --run_count 2
```

guoshzhao: I think it's just one example to show how to use, not for user to use directly, right?

abuccts (approved):

PR #61, opened 2021-04-20
CI/CD - Speedup CPU pipeline

abuccts:
__Major Revisions__

* Speedup Azure pipeline for CPU unit test.


cp5555 (approved):

.azure-pipelines/cpu-unit-test.yml
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) Microsoft Corporation.
```
cp5555, line 1: Just curious: Why can this PR speedup CPU pipeline? Because we remove the docker build time?

.azure-pipelines/cpu-unit-test.yml
```diff
@@ -0,0 +1,27 @@
...
+  - script: |
+      python3 -m pip install .[test,torch]
+    displayName: Install dependencies
+  - script: |
+      python3 setup.py lint
+    displayName: Run code lint
+  - script: |
+      python3 setup.py test
+    displayName: Run unit tests
+    timeoutInMinutes: 10
```
guoshzhao, line 20: timeout will triger failure?

cp5555: I think it will trigger that this test will not pass.

.azure-pipelines/cpu-unit-test.yml
abuccts: yes, it will use pre-built image and skip `apt-get`, etc.

.azure-pipelines/cpu-unit-test.yml
abuccts: yes, it will fail if timeout. this is used to prevent hanging during tests.

guoshzhao (approved):

PR #66, opened 2021-04-27
Benchmarks: Code Revision - Revise MicroBenchmark class to be more flexible.

guoshzhao:
**Description**
Revise MicroBenchmark class to be more flexible.

**Major Revision**
- Complete the _benchmark() function to execute the command list.
- Modify the interface of _process_raw_result() function to support parse the raw output of different commands.
- Add more error handlings.


superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -21,8 +22,10 @@ def __init__(self, name, parameters=''):
         """
         super().__init__(name, parameters)
         self._benchmark_type = BenchmarkType.MICRO
+        # Set the binary directory.
+        self._binary_dir = os.getenv('SUPERBENCH_BINARY_DIR', '/usr/local/bin')
         # Command lines to launch the micro-benchmarks.
-        self.__commands = list()
+        self._commands = list()
```
cp5555, line 28: Why do we need to expose the commands?

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -39,14 +42,40 @@ def _preprocess(self):
...
 
-    @abstractmethod
     def _benchmark(self):
         """Implementation for benchmarking.
 
         Return:
             True if run benchmark successfully.
         """
-        pass
+        for command in self._commands:
```
cp5555, line 51: How can I process the results? The results from all _commands will be merged into one?

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: Here is not for exposing, just for inheriting. If we use self.__commands, the derived class can't access it.

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: The raw result parsing is depend on benchmark owner. For example, P2P bandwidth, Kernel Launch Overhead, benchmark owner should parse the raw string in _process_raw_result(). The results from all commands will be parsed and regarded as different metrics. Most benchmarks just have one command.

cp5555: Got it. To my understanding, if we have multiple commands, they will merge into one for the ooutput.stdout. Correct? Therefore, it should do a well design for the _process_raw_result().

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -21,8 +22,10 @@ def __init__(self, name, parameters=''):
         """
         super().__init__(name, parameters)
         self._benchmark_type = BenchmarkType.MICRO
+        # Set the binary directory.
+        self._binary_dir = os.getenv('SUPERBENCH_BINARY_DIR', '/usr/local/bin')
```
abuccts, line 26: what is this used for? `subprocess.run` can execute executables in system path directly. finding the path is what os does

guoshzhao: SUPERBENCH_BINARY_DIR is the environment variable for superbench binary dir, like /opt/superbench/bin, all microbenchmarks will find the binary from the directory.

abuccts: yes those binaries will be placed in one directory, but I mean we can find/run them directly without the prefix
besides, you may run through mpirun/python which won't be placed in superbench directory

guoshzhao: Didn't get your point. Just want to confirm we are on the same page:

**Dependency (binary and library)**
As we discussed, we will handle the third-party projects and ourselves projects in two ways, third-party projects will be put as submodules in superbenchmark/thirdparty, and ourselves project will be put at superbenchmark/superbench/benchmarks/micro_benchmarks/xxx/. Both of them will be constructed using cmake or make, and the the binary will be put at SUPERBENCH_BINARY_DIR, for example /opt/superbench/bin, and library will be put at SUPERBENCH_LIBRARY_DIR, for example /opt/superbench/lib. For this part, I thought we have reach an agreement from last discussion, right?

**Micro-Benchmarks**
All our benchmarks are wrote in python and inherit from the base class "Benchmark", executor only interact with benchmarks through BenchmarkRegistry as we discussed before. Even you run through mpirun, the command executed by mpirun is our python benchmark, not the binary (dependency) itself. I don't think it's a good way for runner or executor itself to have another way to create and maintain some benchmarks. We should manage all benchmarks in a unify way.

We can have more discussion if there have anything unclear. @abuccts @cp5555

abuccts, line 26: rename to `SB_BIN_PATH` or `SB_PATH`?

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -70,21 +71,139 @@ def _process_numeric_result(self, metric, result):
...
+        """Preprocess/preparation operations before the benchmarking.
+
+        Return:
+            True if _preprocess() succeed.
+        """
+        if not super()._preprocess():
+            return False
+
+        # Set the environment path.
+        if 'SB_PATH' in os.environ:
```
cp5555, line 153: It's only for Microbenchmark or general marco? It would be strange if it's a general usage.

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -70,21 +71,139 @@ def _process_numeric_result(self, metric, result):
...
         return True
 
-    def _process_raw_result(self, raw_output):
+    def print_env_info(self):
+        """Print environments or dependencies information."""
+        # TODO: will implement it when add real benchmarks in the future.
+        pass
+
+
+class MicroBenchmarkWithInvoke(MicroBenchmark):
```
cp5555, line 80: Shall we add UT for this class in this PR?

cp5555 (changes requested):
Add UT for the new class.

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: It's only for Microbenchmark, yifan suggested the name and the env is set by his module. Do yo have any suggestions?

abuccts, line 153: not need to check this `if` if use `os.getenv('SB_PATH', '')`?

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -70,21 +71,139 @@ def _process_numeric_result(self, metric, result):
...
+            name (str): benchmark name.
+            parameters (str): benchmark parameters.
+        """
+        super().__init__(name, parameters)
+
+        # Command lines to launch the micro-benchmarks.
+        self._commands = list()
+
+        # Binary name of the current micro-benchmark.
+        self._bin_name = None
```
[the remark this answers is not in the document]
guoshzhao: If there have several executable files, they should be different benchmarks? One benchmark should only have one executable file.
'git clone' belongs to the commands for preparation, not benchmark.

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -70,21 +71,139 @@ def _process_numeric_result(self, metric, result):
...
+    def add_parser_arguments(self):
+        """Add the specified arguments."""
+        super().add_parser_arguments()
+
+        self._parser.add_argument(
+            '--bin_path',
+            type=str,
+            default=None,
+            required=False,
+            help='Specify the binary path of the benchmark.',
```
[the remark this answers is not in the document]
guoshzhao: Have changed to 

> help='Specify the path of the benchmark binary.'

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -70,21 +71,139 @@ def _process_numeric_result(self, metric, result):
...
+        """
+        if self._bin_name is None:
+            self._result.set_return_code(ReturnCode.MICROBENCHMARK_BINARY_NAME_NOT_SET)
+            logger.error('The binary name is not set - benchmark: {}.'.format(self._name))
+            return False
+
+        existed = True
+        if self._args.bin_path is None:
+            self._args.bin_path = shutil.which(self._bin_name)
+            print(shutil.which(self._bin_name))
```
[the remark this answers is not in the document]
guoshzhao: Done

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -70,21 +71,139 @@ def _process_numeric_result(self, metric, result):
...
+            True if the binary exists.
+        """
+        if self._bin_name is None:
+            self._result.set_return_code(ReturnCode.MICROBENCHMARK_BINARY_NAME_NOT_SET)
+            logger.error('The binary name is not set - benchmark: {}.'.format(self._name))
+            return False
+
+        existed = True
+        if self._args.bin_path is None:
+            self._args.bin_path = shutil.which(self._bin_name)
```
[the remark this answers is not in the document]
guoshzhao: Yes.

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: It can avoid adding empty string into $PATH, even there has no impact....

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -70,21 +71,139 @@ def _process_numeric_result(self, metric, result):
...
+            print(shutil.which(self._bin_name))
+            existed = (self._args.bin_path is not None)
+        else:
+            self._args.bin_path = os.path.join(self._args.bin_path, self._bin_name)
+            existed = os.path.isfile(self._args.bin_path)
+
+        if not existed:
```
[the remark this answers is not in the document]
guoshzhao: Thanks

superbench/benchmarks/micro_benchmarks/micro_base.py
cp5555: Maybe SB_MICRO_PATH?

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: Is SB_MICRO_PATH OK for you? @abuccts

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: Done.

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: SB_MICRO_PATH

superbench/benchmarks/micro_benchmarks/micro_base.py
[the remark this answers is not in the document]
guoshzhao: directory

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: SB_MICRO_PATH

abuccts (approved):

superbench/benchmarks/micro_benchmarks/micro_base.py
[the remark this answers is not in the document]
guoshzhao: Use bin_dir to indicate where the binary locate.

PR #67, opened 2021-05-06
Utils - Support lazy import

abuccts:
__Major Revision__

* Support lazy import.
* Not importing benchmarks when running `help`, `version`, `deploy` commands, etc.

### DIFF
--- a/superbench/benchmarks/__init__.py
+++ b/superbench/benchmarks/__init__.py
@@ -3,10 +3,22 @@
 
 """Exposes interfaces of benchmarks used by SuperBench executor."""
 
+import importlib
+
 from superbench.benchmarks.return_code import ReturnCode
 from superbench.benchmarks.context import Platform, Framework, Precision, ModelAction, BenchmarkType, BenchmarkContext
-from superbench.benchmarks.registry import BenchmarkRegistry
-from superbench.benchmarks import model_benchmarks, micro_benchmarks, docker_benchmarks    # noqa pylint: disable=unused-import
+from superbench.common.utils import LazyImport
+
+BenchmarkRegistry = LazyImport(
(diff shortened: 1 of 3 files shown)


superbench/benchmarks/__init__.py
```diff
@@ -3,10 +3,22 @@
...
+import importlib
+
 from superbench.benchmarks.return_code import ReturnCode
 from superbench.benchmarks.context import Platform, Framework, Precision, ModelAction, BenchmarkType, BenchmarkContext
-from superbench.benchmarks.registry import BenchmarkRegistry
-from superbench.benchmarks import model_benchmarks, micro_benchmarks, docker_benchmarks    # noqa pylint: disable=unused-import
+from superbench.common.utils import LazyImport
+
+BenchmarkRegistry = LazyImport(
+    'superbench.benchmarks.registry', 'BenchmarkRegistry', lambda: list(
```
cp5555, line 13: why do we need superbench.benchmarks.registry?

superbench/common/utils/lazy_import.py
```diff
@@ -0,0 +1,52 @@
...
+        Args:
+            item (str): Attribute name.
+
+        Returns:
+            Any: Attribute value.
+        """
+        self._import()
+        return getattr(self._module, item)
+
+    def __dir__(self):
```
cp5555, line 45: Are there any other functions that we need to re-write it?

cp5555: According to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/util/lazy_loader.py, __getattr__ and __dir__ should be enough now.

cp5555 (approved):

guoshzhao (approved):

PR #70, opened 2021-05-10
CI/CD - Add GitHub Action to build and push image

abuccts:
__Major Revision__

* add GitHub Action to build and push image

__Minor Revision__

* update Dockerfile to copy from context


cp5555 (approved):

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,48 @@
+name: Build Image
```
cp5555, line 1: When does GitHub Action execute?

abuccts: build image and push to Docker Hub when push to main. build image in pull request, 
```yaml
on:
  push:
    branches:
    - main
  pull_request:
    branches:
    - main
```

dockerfile/cuda11.1.1.dockerfile
```diff
@@ -83,9 +83,8 @@ ENV PATH="${PATH}:/usr/local/cmake/bin:/usr/local/nccl-tests/build" \
...
-ARG SB_VERSION="main"
-RUN git clone -b ${SB_VERSION} https://github.com/microsoft/superbenchmark ${SB_HOME} && \
-    cd ${SB_HOME} && \
-    python3 -m pip install .[torch]
-
 WORKDIR ${SB_HOME}
+ADD . .
+
+RUN cd ${SB_HOME} && \
+    python3 -m pip install .[torch]
```
guoshzhao, line 90: Just as ever said, we should only install torchvision and transformers pacakges, and using .[torch] will change the torch version. Can we just remove torch requirement from .[torch] or add a new require term like .[torch_pkg]? Since .[torch] is used for test.dockerfile which has no torch environment.

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,48 @@
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          platforms: linux/amd64
+          file: dockerfile/cuda11.1.1.dockerfile
+          push: ${{ github.event_name == 'push' }}
+          tags: superbench/superbench:latest,superbench/superbench:dev-cuda11.1.1
```
cp5555, line 40: Remove the superbench/superbench:dev-cuda11.1.1

guoshzhao (approved):

PR #72, opened 2021-05-12
Benchmarks: Add Feature - Add script to build all cmake benchmark projects.

guoshzhao:
**Description**
Build all cmake benchmark projects..

**Major Revision**
- Loop all directories inside micro_benchmarks folder and check whether there exist CMakeLists.txt to build the benchmark.
- Use $SB_PATH as the target directory.


cp5555 (approved):

superbench/benchmarks/micro_benchmarks/build.sh
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Microsoft Corporation - All rights reserved
+# Licensed under the MIT License
+
+ #!/bin/bash
```
abuccts, line 4: shebang does not work if it's not at the beginning of a file, `#!` must be the first two chars

superbench/benchmarks/micro_benchmarks/build.sh
```diff
@@ -0,0 +1,17 @@
...
+
+SB_PATH="${SB_PATH:-/usr/local/bin}"
+SB_LIB="${SB_LIB:-/usr/local/lib}"
+
+for dir in */ ; do
+    if [ -f $dir/CMakeLists.txt ]; then
```
abuccts, line 10: why not write cmake file for this? e.g.,
```cmake
SB_PATH ?= /usr/local/bin

FILE(GLOB subdirs /*)
FOREACH(subdir ${subdirs})
  # cmake for each subdir ...
  ADD_SUBDIRECTORY(${subdir})
ENDFOREACH()
```

superbench/benchmarks/micro_benchmarks/build.sh
guoshzhao: Done.

superbench/benchmarks/micro_benchmarks/build.sh
guoshzhao: Is there any necessary factors that to use cmake for build script? I think shell script is easy for caller to use, and it has more flexibility in the future.

abuccts (approved):

PR #75, opened 2021-05-17
Benchmarks: Code Revision -  revise CNN example settings and revise ModelBenchmark to expose interface of pin memory

yukirora:
**Description**
revise CNN example settings and revise ModelBenchmark to expose interface of pin memory

**Major Revision**
- Expose interface of the option to pin memory in data loader
- Revise default CNN configuration 
 `--batch_size 192 --precision float32 float16 --num_warmup 64 --num_steps 512 --sample_count 8192 --pin_memory`


cp5555 (approved):

guoshzhao (approved):

superbench/benchmarks/model_benchmarks/model_base.py
```diff
@@ -131,6 +131,13 @@ def add_parser_arguments(self):
             help='Disable GPU training.',
         )
 
+        self._parser.add_argument(
+            '--pin_memory',
+            action='store_true',
+            default=False,
```
abuccts, line 137: should we enable `pin_memory` by default?

cp5555: We can discuss whether we need to change it in other PR.

PR #76, opened 2021-05-17
Benchmarks: Build Pipeline - Call build script when setup environment.

guoshzhao:
**Description**
Call build script when setup environment.

**Major Revision**
- Call build script of benchmarks in Makefile, the command is `make cppbuild`
- Add cppbuild command for testing env.
- Add cppbuild command for docker image.

### DIFF
--- a/.azure-pipelines/cuda-unit-test.yml
+++ b/.azure-pipelines/cuda-unit-test.yml
@@ -16,6 +16,9 @@ steps:
       python3 -m pip install .[test,torch]
     displayName: Install dependencies
   - script: |
+      make cppbuild
+    displayName: Build benchmarks
+  - script: |
       python3 setup.py lint
     displayName: Run code lint
   - script: |
--- a/Makefile
+++ b/Makefile
@@ -12,4 +12,4 @@ cppformat:
 	clang-format --verbose -i $(CPPSOURCES)
 
 cppbuild:
-	cd ./superbench/benchmarks/; bash build.sh
+	cd ./superbench/benchmarks/ &&  bash build.sh
--- a/Makefile
+++ b/Makefile
@@ -12,4 +12,4 @@ cppformat:
 	clang-format --verbose -i $(CPPSOURCES)
 
 cppbuild:
-	cd ./superbench/benchmarks/ &&  bash build.sh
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -10,3 +10,6 @@ cpplint:
 
 cppformat:
 	clang-format --verbose -i $(CPPSOURCES)
+
+cppbuild:
+	cd ./superbench/benchmarks/; bash build.sh
```
abuccts, line 15: change `;` to `&&`? why not `./superbench/benchmarks/build.sh` directly?

guoshzhao: Both makefile and build.sh use relative path. If use `./superbench/benchmarks/build.sh`, the script will loop `./superbenchmark/` dir not `./superbenchmark/superbench/benchmarks`

cp5555 (approved):

Makefile
```diff
@@ -10,3 +10,6 @@ cpplint:
 
 cppformat:
 	clang-format --verbose -i $(CPPSOURCES)
+
+cppbuild:
+	cd ./superbench/benchmarks/ &&  bash build.sh
```
abuccts, line 15: two spaces after `&&`

.azure-pipelines/cuda-unit-test.yml
```diff
@@ -15,6 +15,9 @@ steps:
   - script: |
       python3 -m pip install .[test,torch]
     displayName: Install dependencies
+  - script: |
+      make cppbuild
+    displayName: Build benchmarks
```
abuccts, line 20: should we build these in unit test? there's no unit test for cpp code, and build process will be checked in docker build ci.

May be build and run these after there's integration tests?

abuccts (approved):

.azure-pipelines/cuda-unit-test.yml
guoshzhao: The cppbuild inside cuda-unit-test.yml is for python benchmarks, we should check something like 'if binary exists' and 'the output format'.

Makefile
guoshzhao: Thanks

PR #81, opened 2021-05-24
Runner - Support torch.distributed mode in runner

abuccts:
__Major Revisions__

* Support `torch.distributed` mode in runner.


superbench/runner/runner.py
```diff
@@ -58,6 +58,27 @@ def __get_enabled_benchmarks(self):
...
+            exec_command (str): Executor command.
+
+        Return:
+            str: Runner command.
+        """
+        if mode == 'torch.distributed':
+            # TODO: replace with torch.distributed.run in v1.9
+            return (
+                'python3 -m torch.distributed.launch '
+                '--use_env --no_python --nproc_per_node=8 '
```
cp5555, line 75: Is it hard code for `--nproc_per_node`?

cp5555 (approved):

superbench/runner/runner.py
```diff
@@ -58,6 +58,27 @@ def __get_enabled_benchmarks(self):
...
+        """Get runner command for given mode.
+
+        Args:
+            mode (str): Runner mode.
+            exec_command (str): Executor command.
+
+        Return:
+            str: Runner command.
+        """
+        if mode == 'torch.distributed':
```
guoshzhao, line 71: How to distinguish "distributed training" and "local parallel training per single node"?

abuccts: there's no need to distinguish? the only difference is `nnode=1` or `nnode>1`

superbench/runner/runner.py
abuccts: currently yes, will change to `--nproc_per_node=gpu` in torch 1.9, here's [reference](https://github.com/pytorch/pytorch/blob/b4f3a989da375675785754e4b12d2ce757a1ffaa/torch/distributed/run.py#L269-L273)

if we want to support dynamic number, e.g. run distributed training on 4 gpus even if there're 8, we may need to add an option in config file.

cp5555: Yes, we definitely need to support it.

superbench/runner/runner.py
```diff
@@ -58,6 +58,30 @@ def __get_enabled_benchmarks(self):
...
+            # TODO: replace with torch.distributed.run in v1.9
+            # TODO: only supports node_num=1 and node_num=all currently
+            return (
+                'python3 -m torch.distributed.launch '
+                '--use_env --no_python --nproc_per_node={proc_num} '
+                '--nnodes={node_num} --node_rank=$NODE_RANK '
+                '--master_addr=$MASTER_ADDR --master_port=$MASTER_PORT '
+                '{command}'
+            ).format(
+                proc_num=mode.proc_num or 8, node_num=1 if mode.node_num == 1 else '$NNODES', command=exec_command
```
guoshzhao, line 81: Who set $NODE_RANK, $NNODES, $MASTER_ADDR etc.? If node_num=1, master_addr should be `localhost`?

guoshzhao (approved):

superbench/runner/runner.py
abuccts: it's set here:
https://github.com/microsoft/superbenchmark/blob/4063d8ebbd5760416f9e002514beccd070c5a483/superbench/runner/playbooks/check_env.yaml#L24-L28

superbench/runner/runner.py

superbench/runner/runner.py
abuccts: added `proc_num` and `node_num` config

PR #83, opened 2021-05-26
Benchmarks: Build Pipeline - Revise path of installing cmake projects

yukirora:
**Description**
Revise path of installing cmake projects

**Major Revision**
- unify SB_MICRO_PATH  to one base path, bin will be installed in SB_MICRO_PATH/bin , library will be installed in SB_MICRO_PATH/lib

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/micro_base.py
+++ b/superbench/benchmarks/micro_benchmarks/micro_base.py
@@ -147,7 +147,8 @@ class MicroBenchmarkWithInvoke(MicroBenchmark):
         # Set the environment path.
         if os.getenv('SB_MICRO_PATH'):
             os.environ['PATH'] = os.path.join(os.getenv('SB_MICRO_PATH'), 'bin') + os.pathsep + os.getenv('PATH', '')
-            os.environ['PATH'] = os.path.join(os.getenv('SB_MICRO_PATH'), 'lib') + os.pathsep + os.getenv('PATH', '')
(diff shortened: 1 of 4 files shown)


superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -145,8 +145,10 @@ def _preprocess(self):
             return False
 
         # Set the environment path.
-        if 'SB_MICRO_PATH' in os.environ:
-            os.environ['PATH'] = os.getenv('SB_MICRO_PATH', '') + os.pathsep + os.getenv('PATH', '')
+        if os.getenv('SB_MICRO_PATH'):
+            os.environ['PATH'] = os.path.join(os.getenv('SB_MICRO_PATH'), 'bin') + os.pathsep + os.getenv('PATH', '')
+            os.environ['LD_LIBRARY_PATH'] = os.path.join(os.getenv('SB_MICRO_PATH'),
```
guoshzhao, line 150: Have you tested that the library can be found by setting 'LD_LIBRARY_PATH'?

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/micro_base.py
yukirora: It works well on my current workspace, but I'm not sure if it will work will in pipeline. But I need this pr to be merged first, and then I can test and see cublas code's status. If it has bug, I will create a new pr to fix it.

PR #85, opened 2021-05-27
Benchmarks: Build Pipeline - Add cutlass as a submodule and add build logic.

guoshzhao:
**Description**
Add cutlass as submodule of superbench.

**Major Revision**
- Add cutlass as submodule
- Add build script for it.
- Call the build script by `make thirdparty`


cp5555 (approved):

Makefile
```diff
@@ -13,6 +13,7 @@ cppformat:
 
 cppbuild:
 	cd ./superbench/benchmarks/ && bash build.sh
+	cd ./third_party/ && bash build.sh
```
abuccts, line 16: change to a new `thirdparty` target?

.azure-pipelines/cuda-unit-test.yml
```diff
@@ -22,6 +22,9 @@ steps:
   - script: |
       python3 setup.py lint
     displayName: Run code lint
+  - script: |
+      git submodule update --init --recursive
+    displayName: Pull all third-party submodules
```
abuccts, line 27: no need in unit test

third_party/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation - All rights reserved
+# Licensed under the MIT License
+
+
+SB_MICRO_PATH="${SB_MICRO_PATH:-/usr/local}"
+
+# Build cutlass.
+if [ -f cutlass/CMakeLists.txt ]; then
```
abuccts, line 10: could we use `Makefile`? since many dependencies do not use cmake
`Makefile` could provide one target for each dependency, so it's easy to support building selected dependencies

third_party/build.sh
```diff
@@ -0,0 +1,16 @@
...
+    SOURCE_DIR=./cutlass
+    BUILD_ROOT=./cutlass/build
+    mkdir -p $BUILD_ROOT
+    cmake -DCMAKE_INSTALL_BINDIR=$SB_MICRO_PATH/bin -DCMAKE_INSTALL_LIBDIR=$SB_MICRO_PATH/lib -DCMAKE_BUILD_TYPE=Release -S $SOURCE_DIR -B $BUILD_ROOT
+    cmake --build $BUILD_ROOT -j 16 --target install
```
abuccts, line 15: should we copy install commands from readme directly?
currently we only need to build `cutlass_profiler`?

.azure-pipelines/cuda-unit-test.yml
guoshzhao: removed

Makefile
guoshzhao: Done

third_party/build.sh
guoshzhao: Since the building of thirdparty may vary a lot, I just used the most general way that using shell script. Each thirdparty project will have its own building script.

third_party/build.sh
guoshzhao: The CMakefile.txt of cutlass has defined `install` target and allow user to specify where to install the binary and library, we can just use it directly. And it also has defined some switches to control what target to build, for now, only cutlass_profiler will be built. 
If we just use the `build` command from readme, we should add the `install` logic by ourselves.

third_party/build.sh
guoshzhao: Changed to Makefile.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "third_party/cutlass"]
```
cp5555, line 1: Therefore, we use submodule for this?

abuccts (approved):

PR #86, opened 2021-05-27
Executor - Save benchmark results to file

abuccts:
__Major Revision__

* Save benchmark results to json file.

### DIFF
--- a/superbench/executor/executor.py
+++ b/superbench/executor/executor.py
@@ -4,6 +4,7 @@
 """SuperBench Executor."""
 
 import json
+import itertools
 from pathlib import Path
 
 from omegaconf import ListConfig
@@ -110,7 +111,7 @@ class SuperBenchExecutor():
                     logger.info('Executor succeeded in %s.', log_suffix)
                 else:
                     logger.error('Executor failed in %s.', log_suffix)
-                return benchmark.result
+                return json.loads(benchmark.serialized_result)
             else:
                 logger.error('Executor failed in %s, invalid context.', log_suffix)
         except Exception as e:
@@ -118,20 +119,21 @@ class SuperBenchExecutor():
             logger.error('Executor failed in %s.', log_suffix)
         return None
 
-    def __get_benchmark_dir(self, benchmark_name, create=False):
-        """Get output directory for benchmark.
+    def __create_benchmark_dir(self, benchmark_name):
+        """Create output directory for benchmark.
 
         Args:
             benchmark_name (str): Benchmark name.
-            create (bool): Create the directory or not.
-
-        Return:
-            str: Benchmark output directory.
         """
         benchmark_output_dir = Path(self._output_dir, 'benchmarks', benchmark_name)
-        if create:
-            benchmark_output_dir.mkdir(mode=0o755, parents=True, exist_ok=True)
-        return str(benchmark_output_dir)
+        if benchmark_output_dir.is_dir() and any(benchmark_output_dir.iterdir()):
+            logger.warn('Benchmark output directory %s is not empty.', str(benchmark_output_dir))
+            for i in itertools.count(start=1):
+                backup_dir = benchmark_output_dir.with_name('{}.{}'.format(benchmark_name, i))
+                if not backup_dir.is_dir():
+                    benchmark_output_dir.rename(backup_dir)
+                    break
+        benchmark_output_dir.mkdir(mode=0o755, parents=True, exist_ok=True)
 
     def __write_benchmark_results(self, benchmark_name, benchmark_results):
         """Write benchmark results.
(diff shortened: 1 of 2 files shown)


superbench/executor/executor.py
```diff
@@ -106,17 +110,47 @@ def __exec_benchmark(self, context, log_suffix):
...
+        Args:
+            benchmark_name (str): Benchmark name.
+            create (bool): Create the directory or not.
+
+        Return:
+            str: Benchmark output directory.
+        """
+        benchmark_output_dir = Path(self._output_dir, 'benchmarks', benchmark_name)
+        if create:
+            benchmark_output_dir.mkdir(mode=0o755, parents=True, exist_ok=True)
```
cp5555, line 133: Do we need to empty the dir?

superbench/executor/executor.py
```diff
@@ -106,17 +110,47 @@ def __exec_benchmark(self, context, log_suffix):
...
+        return str(benchmark_output_dir)
+
+    def __write_benchmark_results(self, benchmark_name, benchmark_results):
+        """Write benchmark results.
+
+        Args:
+            benchmark_name (str): Benchmark name.
+            benchmark_results (dict): Benchmark results.
+        """
+        with Path(self.__get_benchmark_dir(benchmark_name), 'results.json').open(mode='w') as f:
```
cp5555, line 143: Is it a separate json file?

superbench/executor/executor.py
abuccts: nope? if the same output dir is used, later output will rewrite previous one

superbench/executor/executor.py
abuccts: every benchmark will have a results.json in its directory
we may need a merged json after runner collected all json on all nodes

cp5555: I think one json should be good.

superbench/executor/executor.py
```diff
@@ -106,17 +110,47 @@ def __exec_benchmark(self, context, log_suffix):
...
+        """Get output directory for benchmark.
+
+        Args:
+            benchmark_name (str): Benchmark name.
+            create (bool): Create the directory or not.
+
+        Return:
+            str: Benchmark output directory.
+        """
+        benchmark_output_dir = Path(self._output_dir, 'benchmarks', benchmark_name)
```
guoshzhao, line 131: Each run will have different `self._output_dir`, right?

superbench/executor/executor.py
```diff
@@ -106,17 +110,47 @@ def __exec_benchmark(self, context, log_suffix):
...
+                return benchmark.result
             else:
                 logger.error('Executor failed in %s, invalid context.', log_suffix)
-        except Exception:
+        except Exception as e:
+            logger.error(e)
             logger.error('Executor failed in %s.', log_suffix)
+        return None
+
+    def __get_benchmark_dir(self, benchmark_name, create=False):
```
guoshzhao, line 121: Why need one option for `create`?

superbench/executor/executor.py
```diff
@@ -106,17 +110,47 @@ def __exec_benchmark(self, context, log_suffix):
                     logger.info('Executor succeeded in %s.', log_suffix)
                 else:
                     logger.error('Executor failed in %s.', log_suffix)
+                return benchmark.result
```
guoshzhao, line 113: benchmark.result is only the summary data, we'd better return all info of benchmark such as raw_data and meta_data (start/end time, return code, benchmark name etc.). How about:
`return json.loads(benchmark.serialized_result)`

superbench/executor/executor.py
cp5555: Besides, based on our previous bad experience, if we do the merge offline. We don't need to do this in this version. We should do it in the runner.

superbench/executor/executor.py
abuccts: added directory rotation and keep it empty

superbench/executor/executor.py
abuccts: will add in runner in another pr

superbench/executor/executor.py
abuccts: yes, it's create here
https://github.com/microsoft/superbenchmark/blob/f1327bb788aeda7d755ee102e751a8b0edb038fa/superbench/common/utils/file_handler.py#L20

superbench/executor/executor.py
abuccts: removed

superbench/executor/executor.py

cp5555 (approved):

tests/executor/test_executor.py
```diff
@@ -74,6 +75,43 @@ def test_get_arguments(self):
...
+        self.executor._SuperBenchExecutor__create_benchmark_dir('foobar')
+        foobar_results = {
+            'sum': 1,
+            'avg': 1.1,
+        }
+        self.executor._SuperBenchExecutor__write_benchmark_results('foobar', foobar_results)
+        self.assertTrue(foobar_results_path.is_file())
+        with foobar_results_path.open(mode='r') as f:
+            self.assertDictEqual(json.load(f), foobar_results)
+
```
cp5555, line 114: Please improve the code coverage.

guoshzhao (approved):

PR #88, opened 2021-05-30
Runner - Support local mode in runner

abuccts:
__Major Revision__

* Support `local` mode in runner.


superbench/config/default.yaml
```diff
@@ -4,6 +4,10 @@ superbench:
   benchmarks:
     matmul:
       enable: true
+      modes:
+        - name: local
+          proc_num: 8
```
cp5555, line 9: What's the meaning of proc_num? It's not clear for user.

tests/runner/test_runner.py
```diff
@@ -42,6 +42,23 @@ def test_get_mode_command(self):
...
+                },
+                'exec_command': 'sb exec',
+                'expected_command': 'sb exec',
+            },
+            {
+                'mode': {
+                    'name': 'local',
+                    'proc_num': 8,
+                    'proc_rank': 6,
+                    'prefix': 'NVIDIA_VISIBLE_DEVICES={proc_rank} numactl -c $(({proc_rank}/2))'
```
cp5555, line 57: The hardcode is very limited to specific platform. Would you please design a more flexible way?

superbench/runner/runner.py
```diff
@@ -68,7 +69,12 @@ def __get_mode_command(self, mode, exec_command):
         Return:
             str: Runner command.
         """
-        if mode.name == 'torch.distributed':
+        if mode.name == 'local':
+            return '{prefix} {command}'.format(
+                prefix=(mode.prefix or '').format(proc_rank=mode.proc_rank, proc_num=mode.proc_num or 1),
```
cp5555, line 74: Shall we move this logic before return? I think it's hard to debug if there is failure.

superbench/config/default.yaml
abuccts: process number

superbench/runner/runner.py
abuccts: updated

tests/runner/test_runner.py
abuccts: added another case

superbench/config/default.yaml
```diff
@@ -4,6 +4,10 @@ superbench:
...
+          prefix: NVIDIA_VISIBLE_DEVICES={proc_rank}
```
cp5555, line 10: Shall we change to CUDA_VISIBLE_DEVICES?

superbench/runner/runner.py
```diff
@@ -121,7 +128,35 @@ def run(self):
...
+                                    )
+                                ),
+                                sudo=True
+                            )
+                        )
+                        rets[vars['proc_rank']] = rc
+
+                    jobs = []
+                    rets = multiprocessing.Manager().dict()
+                    for proc_rank in range(mode.proc_num or 1):
```
cp5555, line 151: What happens if mode.proc_num is -1? Or 1.2?

superbench/config/default.yaml
abuccts: updated

superbench/runner/runner.py
abuccts: it may fail, haven't added config validation yet
https://github.com/microsoft/superbenchmark/blob/5e9f948df21422f119b6d7ef097ba7265664055e/superbench/executor/executor.py#L45-L51

cp5555: Maybe we should handle it somewhere and we don't use `mode.proc_num or 1` to handle this scenario.

tests/runner/test_runner.py
```diff
@@ -42,6 +42,33 @@ def test_get_mode_command(self):
                 'exec_command': 'sb exec',
```
cp5555, line 42: Shall we improve the code coverage a little bit?

cp5555 (approved):

superbench/runner/runner.py
```diff
@@ -121,7 +128,35 @@ def run(self):
...
+                        proc = multiprocessing.Process(target=run_proc, args=({'proc_rank': proc_rank}, rets))
```
guoshzhao, line 152: Could we support both parallel and sequential execution? In some cases, parallel execution may have performance interference.

superbench/runner/runner.py
guoshzhao: Why prefix need to set `proc_num=mode.proc_num or 1`?

superbench/runner/runner.py
```diff
@@ -68,10 +69,16 @@ def __get_mode_command(self, mode, exec_command):
         Return:
             str: Runner command.
         """
-        if mode.name == 'torch.distributed':
+        mode_command = exec_command
+        if mode.name == 'local':
+            mode_command = '{prefix} {command}'.format(
+                prefix=(mode.prefix or '').format(proc_rank=mode.proc_rank, proc_num=mode.proc_num or 1),
```
guoshzhao, line 75: Why prefix need to set proc_num=mode.proc_num or 1?

superbench/runner/runner.py
```diff
@@ -121,7 +128,35 @@ def run(self):
...
+                        rets[vars['proc_rank']] = rc
+
+                    jobs = []
+                    rets = multiprocessing.Manager().dict()
+                    for proc_rank in range(mode.proc_num or 1):
+                        proc = multiprocessing.Process(target=run_proc, args=({'proc_rank': proc_rank}, rets))
+                        jobs.append(proc)
+                    for proc in jobs:
+                        proc.start()
+                    for proc in jobs:
```
guoshzhao, line 156: Should to add log to check which job finished, which job may hang?

superbench/runner/runner.py
abuccts: updated

tests/runner/test_runner.py
abuccts: updated

superbench/runner/runner.py

superbench/runner/runner.py
abuccts: it's a format string

superbench/runner/runner.py
abuccts: currently it can be determined in logs, we may reply on ansible timeout for such issues in the future

guoshzhao (approved):

PR #90, opened 2021-06-01
Benchmarks: Code Revision - add error return code for cublas microbenchmark

yukirora:
**Description**
Add error return code for cublas microbenchmark

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/cublas_function.py
+++ b/superbench/benchmarks/micro_benchmarks/cublas_function.py
@@ -8,7 +8,7 @@ import json
 import yaml
 
 from superbench.common.utils import logger
-from superbench.benchmarks import Platform, BenchmarkRegistry
+from superbench.benchmarks import Platform, BenchmarkRegistry, ReturnCode
 from superbench.benchmarks.micro_benchmarks import MicroBenchmarkWithInvoke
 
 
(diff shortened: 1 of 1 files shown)


superbench/benchmarks/micro_benchmarks/cublas_function.py
```diff
@@ -293,13 +293,15 @@ def _process_raw_result(self, cmd_idx, raw_output):
                     self._curr_run_index, cmd_idx, self._name, raw_output, str(e)
                 )
             )
+            self._result.set_return_code(ReturnCode.MICROBENCHMARK_RESULT_PARSING_FAILURE)
```
guoshzhao, line 296: The common error like "RESULT_PARSING_FAILURE" can be handled by the base class:
https://github.com/microsoft/superbenchmark/blob/40d7905e6b7a551488b47ebac74802880e84ae44/superbench/benchmarks/micro_benchmarks/micro_base.py#L188

For _process_raw_result(), if return False, the base class will set return code as "RESULT_PARSING_FAILURE", no matter what you set it inside the function.

I think only the specific error in _preprocess need to be handled.

guoshzhao: Same with another PR.

yukirora: I see, thank you for the comment

guoshzhao (approved):

cp5555 (approved):

PR #93, opened 2021-06-03
Benchmarks: Fix Bug - Fix OOM issue when run pytorch models sequentially.

guoshzhao:
**Description**
Clean up cuda cache after every run of pytorch benchmark.


superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -183,6 +183,8 @@ def _postprocess(self):
             )
             return False
 
+        torch.cuda.empty_cache()
```
cp5555, line 186: Is it possible to add a test for this?

guoshzhao: Thanks, have added tests for it. Please check.

cp5555 (approved):

tests/benchmarks/model_benchmarks/test_pytorch_base.py
```diff
@@ -231,3 +229,34 @@ def test_pytorch_base():
...
+
+    # Test cache empty by manually calling torch.cuda.empty_cache().
+    parameters = '--batch_size 32 --num_warmup 8 --num_steps 64 --model_action train'
+    benchmark = PytorchMNIST('pytorch-mnist', parameters=parameters)
+    assert (benchmark)
+    assert (benchmark._preprocess())
+    assert (benchmark._benchmark())
+    del benchmark
+    assert (torch.cuda.memory_stats()['reserved_bytes.all.current'] > 0)
+    torch.cuda.empty_cache()
```
abuccts, line 252: why not run `benchmark._postprocess()` here?

abuccts (approved):

tests/benchmarks/model_benchmarks/test_pytorch_base.py
guoshzhao: Here, all resources have been cleaned manually:
```
del benchmark
torch.cuda.empty_cache()
```
Don't need to call `_postprocess()` again.

abuccts: I mean this test case is used to check `_postprocess` will cleanup resource, so that before the fix the case would fail, with this fix the case will pass.

guoshzhao: The test function `test_pytorch_empty_cache()` is to test resource cleanup, not only for `_postprocess()`.
It has two sub-cases, one is to test the API `torch.cuda.empty_cache()` which is the one you pointed here. Another is to tested automatic resource cleanup in `launch_benchmark()`.

For `_postprocess()`, it is tested implicitly in the second sub-case.

PR #100, opened 2021-06-21
Benchmarks: Add Feature - Add DistributedImpl and DistributedBackend arguments for micro benchmark.

guoshzhao:
**Description**
For distributed micro-benchmarks, add DistributedImpl and DistributedBackend arguments. Currently the benchmark 'computation_communication_overlap' and 'sharding_matmul' need to add these arguments.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
+++ b/superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
@@ -21,7 +21,7 @@ import time
 import torch
 
 from superbench.common.utils import logger
-from superbench.benchmarks import BenchmarkRegistry, ReturnCode
+from superbench.benchmarks import DistributedImpl, DistributedBackend, BenchmarkRegistry, ReturnCode
 from superbench.benchmarks.micro_benchmarks import MicroBenchmark
 from superbench.benchmarks.context import Enum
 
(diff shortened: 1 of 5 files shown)


cp5555 (approved):

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
```diff
@@ -114,6 +114,21 @@ def add_parser_arguments(self):
             required=False,
             help='The number of test step.',
         )
+        self._parser.add_argument(
+            '--distributed_impl',
+            type=DistributedImpl,
+            default=DistributedImpl.DDP,
```
abuccts, line 120: do we need to set ddp as default for models?
https://github.com/microsoft/superbenchmark/blob/3d72c07807e23d572c0889faa481cfe504e947fd/superbench/benchmarks/model_benchmarks/model_base.py#L111-L114

abuccts (approved):

superbench/benchmarks/micro_benchmarks/computation_communication_overlap.py
guoshzhao: No. Since ddp is the only impl for these microbenchmark, but for model benchmark, the impl can be None which means single GPU.

PR #103, opened 2021-06-24
Benchmarks: Add Configuration - Add validation config file for azure NDv4.

guoshzhao:
**Description**
Modify the default configurations to align with previous testings.


superbench/config/default.yaml
```diff
@@ -19,13 +19,17 @@ superbench:
...
+      num_steps: 2048
+      sample_count: 8192
+      batch_size: 32
       precision:
         - float32
         - float16
       model_action:
         - train
+      distributed_backend: nccl
+      distributed_impl: ddp
```
abuccts, line 32: remove? parameters here are for single process

guoshzhao: So to run parallel training, we need to add arguments into the CLI command?
Maybe I should add a new config file for validation, since we didn't run model with single GPU for validation.

cp5555 (approved):

superbench/config/default.yaml
abuccts: this config section is for single process
to run distributed version, you need to use `sb run` and torch.distributed mode will add these two parameters

guoshzhao: I see. Removed.

abuccts: we may move this hardcode into mode like mode.parameters in the future

superbench/config/azure_NDv4.yaml
```diff
@@ -0,0 +1,100 @@
+# SuperBench Config
```
abuccts, line 1: rename file to `azure_ndv4.yaml`?

abuccts (approved):

superbench/config/azure_NDv4.yaml
guoshzhao: Done.

cp5555 (approved):

PR #108, opened 2021-06-29
CLI - Support host-list for deploy and run commands

abuccts:
Support `--host-list` for deploy and run commands.

Before this change, an inventory file is needed to use `sb deploy/run`.
Now, `--host-list localhost` or `-l localhost` is sufficient for quick try.


cp5555 (approved):

docs/cli.md
```diff
@@ -59,6 +59,11 @@ sb deploy [--docker-image]
 
 #### Examples
 
+Deploy default image on current GPU node:
```
guoshzhao, line 62: 'current GPU node' => 'local node'?

docs/cli.md
```diff
@@ -131,6 +136,11 @@ sb run [--config-file]
 
 #### Examples
 
+Run all benchmarks on current GPU node:
```
guoshzhao, line 139: Same。

superbench/runner/ansible.py
```diff
@@ -29,13 +29,16 @@ def __init__(self, config):
...
+            inventory_list = getattr(config, 'host_list', '').strip(',')
+            if inventory_file or inventory_list:
+                self._config['inventory'] = inventory_file or inventory_list
                 self._config['host_pattern'] = 'all'
-                inventory = InventoryManager(loader=DataLoader(), sources=inventory_file)
+                inventory = InventoryManager(loader=DataLoader(), sources=inventory_file or f'{inventory_list},')
                 host_list = inventory.get_groups_dict()['all']
                 if len(host_list) > 0:
                     self._config['cmdline'] = '--forks {}'.format(len(host_list))
+                if inventory_list in ['localhost', '127.0.0.1']:
```
guoshzhao, line 40: what will happen if the inventory_list is 'localhost,node2,node3'?

docs/cli.md
guoshzhao, line 62: 'current GPU node' => 'specified nodes'? 
'current' here looks a little inexact.

superbench/cli/_help.py
```diff
@@ -32,6 +32,8 @@
     type: command
     short-summary: Deploy the SuperBench environments to all given nodes.
     examples:
+        - name: deploy default image on current GPU node
```
guoshzhao, line 35: same for 'current'.

superbench/cli/_help.py
```diff
@@ -52,6 +54,8 @@
     type: command
     short-summary: Run the SuperBench benchmarks distributedly.
     examples:
+        - name: run all benchmarks on current GPU node
```
guoshzhao, line 57: same.

superbench/runner/ansible.py
abuccts: it will try to connect to three nodes using ssh, for fine-grained config (local connect to localhost, ssh to others), an inventory file is needed

docs/cli.md
abuccts: this is a specific example running on `localhost`, so it only means current node

superbench/cli/_help.py
abuccts: this is a specific example, instead of explaining the usage of argument

guoshzhao: Agree. If it only mean local node, then we can use local instead of current?

superbench/runner/ansible.py
guoshzhao: Do we have comments or description for these kind of cases?

guoshzhao (approved):

TobeyQin (approved):

docs/cli.md
abuccts: updated

docs/cli.md
abuccts: changed to "local GPU node"

superbench/cli/_help.py
abuccts: updated to "local", thanks

superbench/cli/_help.py
abuccts: updated to "local"

superbench/runner/ansible.py
abuccts: nope, the `--host-list` is mainly targeted on localhost test, so you don't need to prepare a config file.
it's recommended to prepare a host config for other cases, so we only document how to prepare config files for those cases, including [mixing localhost and remote](https://microsoft.github.io/superbenchmark/docs/getting-started/configuration#ansible-inventory)

PR #116, opened 2021-07-01
Runner - Fetch benchmarks results on all nodes

abuccts:
Fetch benchmarks results on all nodes, will rsync after each benchmark.
The results directory structure on control node is as follows:

```
outputs/
└── datetime
    ├── nodes
    │   └── node-0
    │       ├── benchmarks
    │       │   ├── benchmark-0
    │       │   │   ├── rank-0
    │       │   │   │   └── results.json
    │       └── sb-exec.log
    ├── sb-run.log
    └── sb.config.yaml
```

### DIFF
--- a/superbench/executor/executor.py
+++ b/superbench/executor/executor.py
@@ -3,6 +3,7 @@
 
 """SuperBench Executor."""
 
+import os
 import json
 import itertools
 from pathlib import Path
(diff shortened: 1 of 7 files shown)


superbench/runner/playbooks/fetch_results.yaml
```diff
@@ -0,0 +1,11 @@
+- name: Fetch Results
+  hosts: all
+  gather_facts: true
+  vars:
+    workspace: '{{ ansible_user_dir }}/sb-workspace'
+  tasks:
+    - name: Synchronize Output Directory
+      ansible.posix.synchronize:
```
abuccts, line 8: add dep

superbench/runner/playbooks/fetch_results.yaml
```diff
@@ -0,0 +1,11 @@
...
+        mode: pull
+        src: '{{ workspace }}/{{ sb_output_dir }}/'
+        dest: '{{ absolute_output_dir }}/nodes/{{ ansible_hostname }}'
```
cp5555, line 11: In future, we should think about uploading all results to storage directly.

superbench/executor/executor.py
```diff
@@ -122,21 +123,34 @@ def __exec_benchmark(self, context, log_suffix):
...
+            logger.warning('Benchmark output root directory %s is not empty.', str(benchmark_output_root_dir))
             for i in itertools.count(start=1):
-                backup_dir = benchmark_output_dir.with_name('{}.{}'.format(benchmark_name, i))
+                backup_dir = benchmark_output_root_dir.with_name('{}.{}'.format(benchmark_name, i))
                 if not backup_dir.is_dir():
-                    benchmark_output_dir.rename(backup_dir)
+                    benchmark_output_root_dir.rename(backup_dir)
                     break
-        benchmark_output_dir.mkdir(mode=0o755, parents=True, exist_ok=True)
+        self.__get_benchmark_dir(benchmark_name).mkdir(mode=0o755, parents=True, exist_ok=True)
```
cp5555, line 153: I think we should consider the failure if we cannot create or open this dir.

superbench/runner/runner.py
```diff
@@ -156,6 +157,19 @@ def check_env(self):    # pragma: no cover
             )
         )
 
+    def fetch_results(self):    # pragma: no cover
+        """Fetch benchmark results on all nodes."""
+        (self._output_path / 'nodes').mkdir(mode=0o755, parents=True, exist_ok=True)
```
cp5555, line 162: For example, the disk is full, how to handle it.

superbench/executor/executor.py
```diff
@@ -122,21 +123,38 @@ def __exec_benchmark(self, context, log_suffix):
...
         return None
 
+    def __get_benchmark_dir(self, benchmark_name):
+        """Get output directory for benchmark's current rank.
+
+        Args:
+            benchmark_name (str): Benchmark name.
+        """
+        benchmark_output_dir = self._output_path / 'benchmarks' / benchmark_name
+        for rank_env in ['PROC_RANK', 'LOCAL_RANK']:
```
guoshzhao, line 133: Can we say rank value match the gpu id for both micro and model benchmark? In other words, how can we match the log to hardware?

superbench/executor/executor.py
```diff
@@ -122,21 +123,38 @@ def __exec_benchmark(self, context, log_suffix):
...
-        benchmark_output_dir.mkdir(mode=0o755, parents=True, exist_ok=True)
+        benchmark_output_root_dir = self._output_path / 'benchmarks' / benchmark_name
+        try:
+            if benchmark_output_root_dir.is_dir() and any(benchmark_output_root_dir.iterdir()):
+                logger.warning('Benchmark output root directory %s is not empty.', str(benchmark_output_root_dir))
+                for i in itertools.count(start=1):
+                    backup_dir = benchmark_output_root_dir.with_name('{}.{}'.format(benchmark_name, i))
+                    if not backup_dir.is_dir():
+                        benchmark_output_root_dir.rename(backup_dir)
+                        break
```
guoshzhao, line 153: Can we make it as a util function?  Looks the input can be root_dir and sub_dir.

cp5555 (approved):

superbench/executor/executor.py
abuccts: it's decided by benchmark and its config, e.g., you can use gpu_id=rank/2, or gpu_id=rank*2, etc. for pytorch ddp, gpu_id=rank.

superbench/executor/executor.py
abuccts: do you mean `__get_benchmark_dir`? the rotation here only handles root dir.

guoshzhao: I mean the function to check the existence of one dir and rename it. We have the similiar function in sb v0.1

guoshzhao (approved):

PR #117, opened 2021-07-02
Bug Fix -  Fix race condition issue for multi ranks

abuccts:
__Major Revisions__
- Fix race condition issue when multi ranks rotating the same directory.


.azure-pipelines/ansible-integration-test.yml
```diff
@@ -3,6 +3,7 @@
```
cp5555, line 3: Separate the PR..

guoshzhao (approved):

cp5555 (approved):

PR #118, opened 2021-07-07
Utils: Code Revision - Update network common utils

yukirora:
**Description**
Update network common utils.

**Major Revision**
- Add get_ib_devices in network common utils 
- Move get_free_port from test utils to network common utils

### DIFF
--- a/superbench/common/utils/__init__.py
+++ b/superbench/common/utils/__init__.py
@@ -10,6 +10,6 @@ from superbench.common.utils.lazy_import import LazyImport
 nv_helper = LazyImport('superbench.common.utils.nvidia_helper')
 
 __all__ = [
-    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'nv_helper', 'rotate_dir',
-    'network'
+    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'network', 'nv_helper',
+    'rotate_dir'
 ]
--- a/superbench/common/utils/__init__.py
+++ b/superbench/common/utils/__init__.py
@@ -10,6 +10,12 @@ from superbench.common.utils.lazy_import import LazyImport
 nv_helper = LazyImport('superbench.common.utils.nvidia_helper')
 
 __all__ = [
-    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'network', 'nv_helper',
-    'rotate_dir'
+    'LazyImport',
+    'SuperBenchLogger',
+    'create_sb_output_dir',
+    'get_sb_config',
+    'logger',
+    'network',
+    'nv_helper',
+    'rotate_dir',
 ]
--- a/superbench/common/utils/__init__.py
+++ b/superbench/common/utils/__init__.py
@@ -10,11 +10,6 @@ from superbench.common.utils.lazy_import import LazyImport
 nv_helper = LazyImport('superbench.common.utils.nvidia_helper')
 
 __all__ = [
-    'LazyImport',
-    'SuperBenchLogger',
-    'create_sb_output_dir',
-    'get_sb_config',
-    'logger',
-    'nv_helper',
-    'rotate_dir',
+    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'nv_helper', 'rotate_dir',
+    'network'
 ]
--- a/superbench/common/utils/network.py
+++ b/superbench/common/utils/network.py
@@ -1,0 +1,41 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Network Utility."""
+
+import socket
+import subprocess
+from contextlib import closing
+
+
+def get_free_port():
+    """Get a free port in current system.
+
+    Return:
+        port (int): a free port in current system.
+    """
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        s.bind(('', 0))
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        return s.getsockname()[1]
+
+
+def get_ib_devices():
+    """Get available IB devices in the system and filter ethernet devices.
+
+    Return:
+        ib_devices (list): IB devices in current system.
+    """
+    # Filter 'InfiniBand' devices by link_layer
+    command = "ibv_devinfo | awk '$1 ~ /hca_id/||/link_layer:/ {print $1,$2}'"
+    output = subprocess.run(
+        command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True, check=False, universal_newlines=True
+    )
+    lines = list(filter(None, output.stdout.replace('\n', ' ').split('hca_id:')))
+    ib_devices = []
+    for line in lines:
+        if 'InfiniBand' in line:
+            ib_devices.append(line.split()[0])
+    ib_devices.sort()
+
+    return ib_devices
--- a/superbench/common/utils/network.py
+++ b/superbench/common/utils/network.py
@@ -5,7 +5,6 @@
 
 import socket
 import subprocess
-from contextlib import closing
 
 
 def get_free_port():
@@ -14,10 +13,15 @@ def get_free_port():
     Return:
         port (int): a free port in current system.
     """
-    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
         s.bind(('', 0))
         s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         return s.getsockname()[1]
+    except OSError:
+        return None
+    finally:
+        s.close()
 
 
 def get_ib_devices():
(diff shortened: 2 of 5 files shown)


superbench/common/utils/network.py
```diff
@@ -0,0 +1,41 @@
...
+    """Get a free port in current system.
+
+    Return:
+        port (int): a free port in current system.
+    """
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        s.bind(('', 0))
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        return s.getsockname()[1]
+
```
cp5555, line 21: What will we return if closing fails?

superbench/common/utils/network.py
```diff
@@ -0,0 +1,41 @@
...
+from contextlib import closing
+
+
+def get_free_port():
+    """Get a free port in current system.
+
+    Return:
+        port (int): a free port in current system.
+    """
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
```
cp5555, line 17: What's the benefit of using `with closing()`?

superbench/common/utils/network.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    # Filter 'InfiniBand' devices by link_layer
+    command = "ibv_devinfo | awk '$1 ~ /hca_id/||/link_layer:/ {print $1,$2}'"
+    output = subprocess.run(
+        command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True, check=False, universal_newlines=True
+    )
+    lines = list(filter(None, output.stdout.replace('\n', ' ').split('hca_id:')))
+    ib_devices = []
+    for line in lines:
+        if 'InfiniBand' in line:
```
cp5555, line 37: I think you should check `transport: InfiniBand`.

superbench/common/utils/network.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_ib_devices():
+    """Get available IB devices in the system and filter ethernet devices.
+
+    Return:
+        ib_devices (list): IB devices in current system.
+    """
+    # Filter 'InfiniBand' devices by link_layer
+    command = "ibv_devinfo | awk '$1 ~ /hca_id/||/link_layer:/ {print $1,$2}'"
```
cp5555, line 30: Why not `ibv_devinfo` directly?

superbench/common/utils/network.py
```diff
@@ -0,0 +1,41 @@
...
+    output = subprocess.run(
+        command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True, check=False, universal_newlines=True
+    )
```
cp5555, line 33: What happens if we don't install `ibv_devinfo`?

superbench/common/utils/__init__.py
```diff
@@ -10,11 +10,6 @@
...
 __all__ = [
-    'LazyImport',
-    'SuperBenchLogger',
-    'create_sb_output_dir',
-    'get_sb_config',
-    'logger',
-    'nv_helper',
-    'rotate_dir',
+    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'nv_helper', 'rotate_dir',
+    'network'
```
abuccts, line 14: add trailing comma
sort in alphabet order

superbench/common/utils/network.py
abuccts, line 30: check `/sys/class/infiniband/<device>/ports/<port-num>/link_layer` is `InifiniBand` and get `<device>` list?

reference: [sysfs-class-infiniband](https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-infiniband)

abuccts: could also check `/sys/class/infiniband/<device>/node_type` for transport type
referemce: [dev->transport_type](https://github.com/linux-rdma/rdma-core/blob/d7a9dde2b164a1563c7a51e4ade8ea3b6cdea9ef/libibverbs/init.c#L380-L401)

superbench/common/utils/__init__.py
```diff
@@ -10,11 +10,6 @@
...
 
 __all__ = [
-    'LazyImport',
-    'SuperBenchLogger',
-    'create_sb_output_dir',
-    'get_sb_config',
-    'logger',
-    'nv_helper',
-    'rotate_dir',
+    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'nv_helper', 'rotate_dir',
```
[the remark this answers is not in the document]
yukirora: It was automatically changed by lint and format, not me

superbench/common/utils/network.py
yukirora: change to try-catch style, but close() was designed to put in finally

superbench/common/utils/network.py
yukirora: it's an error-handling for socket exception

superbench/common/utils/network.py
yukirora: transport can not differentiate

superbench/common/utils/network.py
yukirora: change to another check interface according to Yifan's comments

superbench/common/utils/network.py
yukirora: change to another check interface according to Yifan's comments

superbench/common/utils/__init__.py
yukirora: thank you

superbench/common/utils/__init__.py
```diff
@@ -10,6 +10,6 @@
 nv_helper = LazyImport('superbench.common.utils.nvidia_helper')
 
 __all__ = [
-    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'nv_helper', 'rotate_dir',
-    'network'
+    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'network', 'nv_helper',
+    'rotate_dir'
```
cp5555, line 14: one item for one line?

cp5555 (approved):

superbench/common/utils/__init__.py
guoshzhao: For benchmarks submodule, the items are placed in one line.

guoshzhao (approved):

superbench/common/utils/__init__.py
abuccts: it will exceed max line length

superbench/common/utils/__init__.py
```diff
@@ -10,11 +10,6 @@
...
 __all__ = [
-    'LazyImport',
-    'SuperBenchLogger',
-    'create_sb_output_dir',
-    'get_sb_config',
-    'logger',
-    'nv_helper',
-    'rotate_dir',
+    'LazyImport', 'SuperBenchLogger', 'create_sb_output_dir', 'get_sb_config', 'logger', 'network', 'nv_helper',
+    'rotate_dir'
```
abuccts, line 14:
```suggestion
    'rotate_dir',
```

superbench/common/utils/network.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def get_ib_devices():
+    """Get available IB devices in the system and filter ethernet devices.
+
+    Return:
+        ib_devices (list): IB devices in current system.
+    """
+    command_get_devices = 'ls /sys/class/infiniband/'
+    output = subprocess.run(
```
abuccts, line 34: use file operations instead of subprocess?

superbench/common/utils/__init__.py
guoshzhao: Why we should concern about this rule? New line can be created if exceed the max-line-length, right? Please refer:

superbench/benchmarks/micro_benchmarks/__init__.py

superbench/common/utils/network.py
```diff
@@ -0,0 +1,71 @@
...
+        command_get_devices,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        shell=True,
+        check=False,
+        universal_newlines=True
+    )
+    devices = output.stdout.split()
+    devices.sort()
```
abuccts, line 43: use natural sort order? e.g., what if there are 16 devices mlx5_0, mlx5_1, ..., mlx5_15
```suggestion
    devices.sort(key=lambda s: [int(ch) if ch.isdigit() else ch for ch in re.split(r'(\d+)', s)])
```

superbench/common/utils/network.py
```diff
@@ -0,0 +1,71 @@
...
+            output = subprocess.run(
+                command_get_link_layer,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                shell=True,
+                check=False,
+                universal_newlines=True
+            )
+            if 'InfiniBand' in output.stdout:
+                ib_devices.append(device)
```
abuccts, line 68: what if one hca has two ports? here will append it twice

superbench/common/utils/__init__.py
yukirora: thank you for the suggestions

superbench/common/utils/network.py
yukirora: ok, thank you very much

superbench/common/utils/network.py
yukirora: thanks for the comment

superbench/common/utils/network.py
```diff
@@ -0,0 +1,52 @@
...
+        return None
+    finally:
+        s.close()
+
+
+def get_ib_devices():
+    """Get available IB devices with available ports in the system and filter ethernet devices.
+
+    Return:
+        ib_devices (list): IB devices with available ports in current system.
```
abuccts, line 32: update accordingly?

abuccts (approved):

superbench/common/utils/network.py
yukirora: updated

superbench/common/utils/network.py
yukirora: thank you very much

superbench/common/utils/__init__.py
abuccts: this list is updated before when it cannot be fit in one line.
isn't it more clear to have one item per line (adding a trailing comma for last item could achieve this) and sort in alphabet order if the list is too long? e.g., [when it has 50+ items](https://github.com/pytorch/pytorch/blob/v1.9.0/torch/distributions/__init__.py#L118)

guoshzhao: If you have these kind of concerns, we should make a rule to let all list of item to be placed in format 'one item per line', including function arguments, list, tupe and so on.
For __all__ in __init__.py, it's just to control the exporting, not for human readable.

https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L32

abuccts: different module in one project may have different styles, so nobody uses a lint rule to force `__all__`
https://github.com/python/cpython/blob/3.10/Lib/concurrent/futures/__init__.py#L20

PR #119, opened 2021-07-07
Runner & Executor - Support AMD GPU

abuccts:
__Description__
Support both NVIDIA and AMD GPU and check GPU vendor during deployment and execution.

__Major Revisions__
* Add GPU environment check in `sb deploy`
* Check GPU vendor in executor.

### DIFF
--- a/superbench/common/devices/__init__.py
+++ b/superbench/common/devices/__init__.py
@@ -1,0 +1,10 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""SuperBench devices module."""
+
+from superbench.common.devices.gpu import GPU
+
+__all__ = [
+    'GPU',
+]
--- a/superbench/common/devices/gpu.py
+++ b/superbench/common/devices/gpu.py
@@ -1,0 +1,39 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""GPU device module."""
+
+from pathlib import Path
+
+from superbench.benchmarks import Platform
+from superbench.common.utils import logger
+
+
+class GPU():
+    """GPU device helper class."""
+    def __init__(self):
+        """Initilize."""
+        self.vendor = self.get_vendor()
+        # TODO: check CUDA or ROCm availablity accordingly
+        if self.vendor == 'nvidia':
+            self.platform = Platform.CUDA
+        elif self.vendor == 'amd':
+            self.platform == Platform.ROCM
+        else:
+            self.platform = Platform.CPU
+
+    def get_vendor(self):
+        """Get GPU vendor.
+
+        Returns:
+            str: GPU vendor, nvidia or amd.
+        """
+        if Path('/dev/nvidiactl').is_char_device() and Path('/dev/nvidia-uvm').is_char_device():
+            if not list(Path('/dev').glob('nvidia[0-9]*')):
+                logger.warning('Cannot find NVIDIA GPU device.')
+            return 'nvidia'
+        if Path('/dev/kfd').is_char_device() and Path('/dev/dri').is_dir():
+            if not list(Path('/dev/dri')).glob('card*'):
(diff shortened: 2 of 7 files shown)


superbench/runner/playbooks/deploy.yaml
```diff
@@ -25,16 +25,26 @@
...
     - name: Set GPU Facts
       set_fact:
-        nvidia_gpu_exist: "{{ nvidia_dev.stat.ischr is defined and nvidia_dev.stat.ischr }}"
-        amd_gpu_exist: "{{ amd_dev.stat.ischr is defined and amd_dev.stat.ischr }}"
+        nvidia_gpu_exist: >-
+          {{ nvidia_dev.results[0].stat.ischr is defined and nvidia_dev.results[0].stat.ischr and
+          nvidia_dev.results[1].stat.ischr is defined and nvidia_dev.results[1].stat.ischr }}
+        amd_gpu_exist: >-
+          {{ amd_dev.results[0].stat.ischr is defined and amd_dev.results[0].stat.ischr and
+          amd_dev.results[1].stat.isdir is defined and amd_dev.results[1].stat.isdir }}
```
cp5555, line 47: Would you please write some comments to explain it?

superbench/runner/playbooks/deploy.yaml
```diff
@@ -25,16 +25,26 @@
   tasks:
     - name: Checking NVIDIA GPU
```
cp5555, line 26: It's not check NVIDIA GPU, right? If you would like to check, you should use lspci.

superbench/common/devices/__init__.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""SuperBench devices module."""
+
+from superbench.common.devices.gpu import GPU
+
+__all__ = [
+    'GPU',
+]
```
cp5555, line 10: Why do we put \_\_init\_\_.py in common directly. Currently, we reduce it from `superbench.common.devices.gpu.GPU` to `superbench.common.devices.GPU`

cp5555: Besides, why not put into `utils`?

cp5555:
Please add `__Major Revision__`

superbench/common/devices/gpu.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self):
+        """Initilize."""
+        self.vendor = self.get_vendor()
+        # TODO: check CUDA or ROCm availability accordingly
+        if self.vendor == 'nvidia':
+            self.platform = Platform.CUDA
+        elif self.vendor == 'amd':
+            self.platform == Platform.ROCM
+        else:
+            self.platform = Platform.CPU
```
guoshzhao, line 23: `self.vendor` => `vendor`
`self.platform` => `self.__platform`
Then using @property to get platform?
```
@property
def platform(self):
    return self.__platform
```

superbench/executor/executor.py
```diff
@@ -67,8 +68,7 @@ def __get_enabled_benchmarks(self):
 
     def __get_platform(self):
         """Detect runninng platform by environment."""
-        # TODO: check devices and env vars
-        return Platform.CUDA
+        return GPU().platform
```
guoshzhao, line 71: If we use the structure:
-devices
   --gpu
   --others
how __get_platform() to know devices? Here looks like a hardcode.

superbench/runner/playbooks/deploy.yaml
abuccts: renamed

superbench/common/devices/gpu.py
abuccts: updated

superbench/executor/executor.py

cp5555 (approved):

guoshzhao (approved):

PR #139, opened 2021-07-28
Benchmarks: Build Pipeline - add rccl-tests as a submodule with building logic

yukirora:
**Description**
Support rocm in third_party/makefile and add rccl-tests as a submodule with building logic.

**Major Revision**
- Support rocm in third_party/makefile
- Add rccl-tests as a submodule 
- Add build logic in third_party/Makefile for rccl-tests

### DIFF
--- a/third_party/Makefile
+++ b/third_party/Makefile
@@ -12,7 +12,7 @@ USE_ROCM ?= "0"
 
 .PHONY: all common cuda rocm cutlass bandwidthTest nccl_tests perftest fio rccl_tests
 
-# Build all targets for cuda.
+# Build all targets.
 all: cuda rocm
 cuda: common cutlass bandwidthTest nccl_tests
 rocm: common rccl_tests
--- a/third_party/Makefile
+++ b/third_party/Makefile
@@ -8,13 +8,10 @@ MPI_HOME_PATH ?= "/usr/local/mpi"
 HIP_HOME ?= "/opt/rocm/hip"
 RCCL_HOME ?= "/opt/rocm/rccl"
 
-.PHONY: all common cuda rocm cutlass bandwidthTest nccl_tests perftest fio rccl_tests
+.PHONY: all cutlass bandwidthTest nccl_tests perftest fio
 
 # Build all targets.
-all: cuda rocm
-cuda: common cutlass bandwidthTest nccl_tests perftest_cuda
-rocm: common rccl_tests perftest_rocm
-common: fio
+all: cutlass bandwidthTest nccl_tests perftest fio
 
 # Create $(SB_MICRO_PATH)/bin and $(SB_MICRO_PATH)/lib, no error if existing, make parent directories as needed.
 sb_micro_path:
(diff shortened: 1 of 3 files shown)


third_party/Makefile
```diff
@@ -5,11 +5,16 @@
...
+HIP_HOME ?= "/opt/rocm/hip"
+RCCL_HOME ?= "/opt/rocm/rccl"
 
-.PHONY: all cutlass bandwidthTest nccl_tests perftest fio
+.PHONY: all rocm_all cutlass bandwidthTest nccl_tests perftest fio rccl_tests
 
-# Build all targets.
-all: cutlass bandwidthTest nccl_tests perftest fio
+# Build all targets for cuda.
+all: cutlass bandwidthTest nccl_tests perftest
```
cp5555, line 14: Why remove fio?

cp5555, line 14: I think we would have `make cuda`, `make rocm`, and `make all`.

`make all` should be `all: cuda rocm`

third_party/Makefile
```diff
@@ -5,11 +5,16 @@
 
 SB_MICRO_PATH ?= "/usr/local"
 MPI_HOME_PATH ?= "/usr/local/mpi"
+HIP_HOME ?= "/opt/rocm/hip"
+RCCL_HOME ?= "/opt/rocm/rccl"
 
-.PHONY: all cutlass bandwidthTest nccl_tests perftest fio
+.PHONY: all common cuda rocm cutlass bandwidthTest nccl_tests perftest fio rccl_tests
```
cp5555, line 11: Do you need to revise perftest?

third_party/Makefile
```diff
@@ -5,11 +5,16 @@
```
cp5555, line 5: Shall we separate the PR?
One is only for `Benchmarks: Build Pipeline - Support rocm in third_party/makefile` and the other is `Benchmarks: Build Pipeline - Add rccl-tests as a submodule with building logic`

third_party/Makefile
```diff
@@ -5,11 +5,16 @@
...
 MPI_HOME_PATH ?= "/usr/local/mpi"
+HIP_HOME ?= "/opt/rocm/hip"
+RCCL_HOME ?= "/opt/rocm/rccl"
 
-.PHONY: all cutlass bandwidthTest nccl_tests perftest fio
+.PHONY: all common cuda rocm cutlass bandwidthTest nccl_tests perftest fio rccl_tests
 
 # Build all targets.
-all: cutlass bandwidthTest nccl_tests perftest fio
+all: cuda rocm
```
hpourreza, line 14: wouldn't it better to add common to this line so when someone runs only rocm or cuda, does not run fio?

third_party/Makefile
```diff
@@ -5,6 +5,8 @@
 
 SB_MICRO_PATH ?= "/usr/local"
 MPI_HOME_PATH ?= "/usr/local/mpi"
```
abuccts, line 7: rename to `MPI_HOME`? and better to [remove quotes](https://stackoverflow.com/a/23332194/7630831)
```suggestion
MPI_HOME ?= /usr/local/mpi
```

abuccts (approved):

guoshzhao (approved):

third_party/Makefile
cp5555: For rocm or cuda, we would like to call `make cuda` for running SuperBench in CUDA and call `make rocm` for running SuperBench in ROCM. We think there is very limited case to run `make all`. In future, we may need to add `make cpu`.

third_party/Makefile
yukirora: ok, rename to MPI_HOME and remove all quotes of path in vars

third_party/Makefile
yukirora: ok, thanks for the comments

third_party/Makefile
yukirora: sure

cp5555 (approved):

third_party/Makefile
hpourreza: Does running `make cuda` or `make rocm` in this case, run fio as well? If yes, is it expected?

cp5555: Yes, it will install the FIO because make cuda would be running on the CUDA docker. Currently, if we have a cuda docker and we run `make all`, it will be crash. That's why we design `make cuda` and `make rocm`.

PR #146, opened 2021-08-01
Runner - Support mpi mode

abuccts:
__Description__

Support mpi mode.

__Major Revisions__

* concate mpirun command
* support mca and env config
* prepare hostfile and update Ansible host pattern


superbench/runner/ansible.py
```diff
@@ -78,6 +78,20 @@ def run(self, ansible_config, sudo=False):    # pragma: no cover
...
+    def update_mpi_config(self, ansible_config):
+        """Update ansible config for mpi.
+
+        Args:
+            ansible_config (dict): Ansible config dict.
+
+        Returns:
+            dict: Updated Ansible config dict.
+        """
+        if ansible_config['host_pattern'] != 'localhost':
```
cp5555, line 90: Why is `localhost` different from others? For example, I give 127.0.0.1. It should be the same, correct?

superbench/runner/runner.py
```diff
@@ -48,7 +48,7 @@ def __set_logger(self, filename):
         """
         SuperBenchLogger.add_handler(logger.logger, filename=str(self._output_path / filename))
 
-    def __validate_sb_config(self):
+    def __validate_sb_config(self):    # noqa: C901
```
cp5555, line 51: Why ignore flake8 warning?

superbench/runner/templates/ssh_config.j2
```diff
@@ -1,5 +1,5 @@
 {% for host in hostvars.values() | map(attribute='ansible_hostname') | sort %}
-Host node{{ loop.index0 }}
+Host node{{ loop.index0 }} {{ host }}
```
cp5555, line 2: Why add this?

superbench/runner/runner.py
```diff
@@ -211,6 +234,6 @@ def run(self):
                             'proc_rank': proc_rank
                         }) for proc_rank in range(mode.proc_num)
                     )
-                elif mode.name == 'torch.distributed':
+                elif mode.name == 'torch.distributed' or mode.name == 'mpi':
```
cp5555, line 237: Is our mode.name an enum variable? Here is no `else` to handle the failure scenario.

superbench/runner/runner.py
```diff
@@ -122,6 +134,17 @@ def __get_mode_command(self, benchmark_name, mode):
                     'superbench.benchmarks.{name}.parameters.distributed_backend=nccl'
                 ).format(name=benchmark_name),
             )
+        elif mode.name == 'mpi':
+            mode_command = (
+                'mpirun '
+                '--allow-run-as-root -hostfile hostfile -map-by ppr:{proc_num}:numa '
```
cp5555, line 140: What happens if docker doesn't support `-allow-run-as-root`?

cp5555, line 140: What's meaning of `-map-by`? Would you please all comments to describe the option meaning in the code including `--alow-run-as-root`, `-hostfile`, `-map-by`, etc,?

superbench/runner/runner.py
```diff
@@ -122,6 +134,17 @@ def __get_mode_command(self, benchmark_name, mode):
...
+                '{mca_list} {env_list} {command}'
+            ).format(
+                proc_num=mode.proc_num,
+                mca_list=' '.join(f'-mca {k} {v}' for k, v in mode.mca.items()),
+                env_list=' '.join(f'-x {k}={v}' if v else f'-x {k}' for k, v in mode.env.items()),
+                command=exec_command,
+            )
```
cp5555, line 147: If mode_command doesn't belong to any support feature, it will return `exec_command`. Is it our design?

cp5555 (changes requested):

superbench/runner/templates/ssh_config.j2
abuccts: updated

superbench/runner/runner.py
abuccts: added warning in else

superbench/runner/runner.py

superbench/runner/runner.py
abuccts: it will return original command

superbench/runner/runner.py
abuccts: function too complex, will separate validation part in the future

superbench/runner/ansible.py
cp5555: As Yifan mentioned, it's ansible's configuration

superbench/runner/runner.py
cp5555: Shall we fix it this time?

cp5555 (approved):

guoshzhao (approved):

PR #147, opened 2021-08-04
Benchmarks: Add Feature - Add reduce function support for output summary.

guoshzhao:
**Description**
Add reduce function support for output summary.

**Major Revision**
- Add reducer class to maintain all reduce functions.
- Save reduce type of each metric into `BenchmarkResult`
- Fix UT.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/micro_base.py
+++ b/superbench/benchmarks/micro_benchmarks/micro_base.py
@@ -55,6 +55,7 @@ class MicroBenchmark(Benchmark):
         Args:
             metric (str): metric name which is the key.
             result (List[numbers.Number]): numerical result.
+            reduce_type (ReduceType): The type of reduce function.
 
         Return:
             True if result list is not empty.
--- a/superbench/benchmarks/micro_benchmarks/micro_base.py
+++ b/superbench/benchmarks/micro_benchmarks/micro_base.py
@@ -49,7 +49,7 @@ class MicroBenchmark(Benchmark):
         """
         pass
 
-    def _process_numeric_result(self, metric, result):
+    def _process_numeric_result(self, metric, result, reduce_type=None):
         """Function to save the numerical results.
 
         Args:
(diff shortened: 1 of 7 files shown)


superbench/benchmarks/reducer.py
```diff
@@ -0,0 +1,59 @@
...
+    """The Enum class representing different reducer."""
+    AVG = 'avg'
+    MAX = 'max'
+    MIN = 'min'
+    SUM = 'sum'
+
+
+class Reducer:
+    """Reducer class to maintain all reduce functions."""
+    functions: Dict[ReduceType, Callable] = dict()
```
abuccts, line 22: why not write the content here directly?

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -68,7 +68,7 @@ def _process_numeric_result(self, metric, result):
             return False
 
         self._result.add_raw_data(metric, result)
-        self._result.add_result(metric, sum(result) / len(result))
+        self._result.add_result(metric, sum(result) / len(result), reduce_type)
```
abuccts, line 71: built-in `sum` has [approximation issue](https://docs.python.org/3/tutorial/floatingpoint.html), e.g.,
```
>>> sum([0.1, 0.1, 0.1])
0.30000000000000004
```
```suggestion
        self._result.add_result(metric, statistics.mean(result), reduce_type)
```

superbench/benchmarks/reducer.py
guoshzhao: Want to add reduce functions in a unified way by calling `add_reduce_func()`:

```
@Reducer.add_reduce_func(ReduceType.PERCENT99)
def percent_99(items):
    xxx
```

superbench/benchmarks/micro_benchmarks/micro_base.py
guoshzhao: I will correct it in a separate PR. Thanks!

cp5555 (approved):

abuccts (approved):

PR #149, opened 2021-08-05
Benchmarks: Add Feature - Set reduce type for current benchmarks' metrics.

guoshzhao:
**Description**
Set reduce type for current benchmarks' metrics, including model benchmarks and ShardingMatmul.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/sharding_matmul.py
+++ b/superbench/benchmarks/micro_benchmarks/sharding_matmul.py
@@ -21,6 +21,7 @@ from superbench.common.utils import logger
 from superbench.benchmarks import DistributedImpl, DistributedBackend, BenchmarkRegistry, ReturnCode
 from superbench.benchmarks.micro_benchmarks import MicroBenchmark
 from superbench.benchmarks.context import Enum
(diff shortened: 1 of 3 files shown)


superbench/benchmarks/micro_benchmarks/sharding_matmul.py
```diff
@@ -255,7 +256,7 @@ def _benchmark(self):
                 return False
 
             metric = '{}'.format(mode)
-            if not self._process_numeric_result(metric, elapse_times):
+            if not self._process_numeric_result(metric, elapse_times, reduce_type=ReduceType.MAX):
```
cp5555, line 259: Why do we need to change Sharding Matmul?

guoshzhao: It depends on how many metrics do we want to get, for example:
For sharding matmul, I think we should get one metric no matter how many GPUs are used.
For kernel launch, we should get one metric per GPU.

cp5555: Make sense!

cp5555 (approved):

abuccts (approved):

PR #150, opened 2021-08-05
Benchmarks: Doc Revision - Add ReduceType into benchmarks doc

guoshzhao:
**Description**
Add ReduceType description into benchmarks doc.

### DIFF
--- a/docs/design-docs/benchmarks.md
+++ b/docs/design-docs/benchmarks.md
@@ -32,7 +32,7 @@ The structure of `benchmarks` package can be divided into layers from the bottom
 2. Derived classes for all implemented benchmarks, which need to realize all the abstract interfaces. The benchmarks will be registered into `BenchmarkRegistry`.
 3. `BenchmarkRegistry` provides a way of benchmark registration, maintains all the registered benchmarks, and supports benchmark launching by `BenchmarkContext`.
 4. `BenchmarkContext` provides the context to launch one benchmark, including name, parameters, platform(CPU, GPU, etc.), and framework(Pytorch, TF, ONNX, etc.).
-5. `BenchmarkResult` defines the structured results for each benchmark in json format, including name, return_code, start_time, end_time, raw_data, summarized metrics, etc.
+5. `BenchmarkResult` defines the structured results for each benchmark in json format, including name, return_code, start_time, end_time, raw_data, summarized metrics, reduce type, etc.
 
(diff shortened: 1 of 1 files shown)


docs/design-docs/benchmarks.md
```diff
@@ -244,6 +249,12 @@ result = {
             'throughput-inference-float32': [avg_throughput1, ..., avg_throughputN],
             'throughput-inference-float16': [avg_throughput1, ..., avg_throughputN],
     },
```
cp5555, line 251: Do we need to change result as well?

docs/design-docs/benchmarks.md
```diff
@@ -244,6 +249,12 @@ result = {
...
+    'reduce': {
```
cp5555, line 252: I don't find the code for the name. Can we change to `reduce_operator`? `reduce` is a little bit strange.

docs/design-docs/benchmarks.md
```diff
@@ -181,7 +181,7 @@ This chapter will describe the interfaces with the caller (Superbench executor),
 
 The inputs needed by the `benchmarks` package is simple, just the context object of the benchmark want to run:
 
-### Invoke
+#### Invoke
```
cp5555, line 184: Why change this?

cp5555 (changes requested):

docs/design-docs/benchmarks.md
guoshzhao: It is one member of BenchmarkResult, and has been merged in last PR:
https://github.com/microsoft/superbenchmark/blob/main/superbench/benchmarks/result.py#L34

I will submit separated PR to modify it uniformly.

docs/design-docs/benchmarks.md
guoshzhao: It should be `### Inputs`'s `#### Invoke`, just align with `### Outputs`

docs/design-docs/benchmarks.md
guoshzhao: Here the result is from one executor, it don't need to change. `avg `here means the average of every step.

docs/design-docs/benchmarks.md
cp5555: Got it!

docs/design-docs/benchmarks.md
cp5555: I think the result should be min already. We don't have average anymore.

cp5555: Got it. It's for executor output.

cp5555 (approved):

TobeyQin (approved):

PR #153, opened 2021-08-11
Benchmarks: Add Benchmark - Add memory bus bandwidth performance microbenchmark for amd 

yukirora:
**Description**
Add memory bus bandwidth performance microbenchmark for amd.

**Major Revision**
- Add memory bus bandwidth performance microbenchmark for amd.
- Add related example and test file.


superbench/benchmarks/micro_benchmarks/rocm_memory_bw_performance.py
```diff
@@ -0,0 +1,122 @@
+# Copyright (c) Microsoft Corporation.
```
cp5555, line 1: One quick question: I find this code is almost the same as cuda_memory_bw_performance, but the parameters are different, why? Shall we need to merge it directly?

PR #161, opened 2021-08-20
Benchmarks: Revise Benchmark - Add readwrite I/O pattern

yzygitzh:
**Description**
This commit adds readwrite I/O pattern for FIO benchmark. Read/write ratio is fixed at 4:1.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/disk_performance.py
+++ b/superbench/benchmarks/micro_benchmarks/disk_performance.py
@@ -26,7 +26,7 @@ class DiskBenchmark(MicroBenchmarkWithInvoke):
         self._bin_name = 'fio'
 
         self.__io_patterns = ['seq', 'rand']
-        self.__io_types = ['read', 'write', 'rw']
+        self.__io_types = ['read', 'write', 'readwrite']
         self.__rand_block_size = 4 * 1024    # 4KiB
         self.__seq_block_size = 128 * 1024    # 128KiB
         self.__default_iodepth = 64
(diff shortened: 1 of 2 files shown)


superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -26,13 +26,14 @@ def __init__(self, name, parameters=''):
         self._bin_name = 'fio'
 
         self.__io_patterns = ['seq', 'rand']
-        self.__io_types = ['read', 'write']
+        self.__io_types = ['read', 'write', 'rw']
```
cp5555, line 29: Shall we change `rw` to `readwrite`

superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -97,7 +100,7 @@ def add_parser_arguments(self):
                     help='Time in seconds to warm up %s test.' % io_str,
                 )
                 # Disable write tests by default
-                default_runtime = 0 if io_type == 'write' else self.__default_runtime
+                default_runtime = 0 if io_type in ['write', 'rw'] else self.__default_runtime
```
cp5555, line 103: One question: do we need to change the output data? Because there are both read and write performance.

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: rw here is to align with FIO configuration. FIO recognizes rw, randrw and readwrite, but not randreadwrite.

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: The output already contains both read and write data. You can check the test for this benchmark.

cp5555: Great, thanks!

superbench/benchmarks/micro_benchmarks/disk_performance.py
cp5555: It's interface for SuperBench, and we can translate it internally. @guoshzhao and @abuccts what's your thought for this?

abuccts (approved):

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: Changed to `readwrite` to align with our existing style. I prefer not to introduce both `rw` and `readwrite` here because it's not a must, and it might mislead user to think of them as two different functions.

cp5555 (approved):

PR #175, opened 2021-08-30
Benchmarks: Code Revision - Revise metric name generation and default config for disk performance benchmark

yzygitzh:
**Description**
This commit revises disk performance benchmark, including:
1) Add missing benchmark name in default config;
2) Avoid using reserved character ':' in metric name.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/disk_performance.py
+++ b/superbench/benchmarks/micro_benchmarks/disk_performance.py
@@ -204,7 +204,7 @@ class DiskBenchmark(MicroBenchmarkWithInvoke):
                 self._result.add_result('%s_iops' % io_type_prefix, float(iops))
 
                 for lat_unit in lat_units:
-                    if lat_unit in fio_output['jobs'][0][io_type]_
+                    if lat_unit in fio_output['jobs'][0][io_type]:
                         lat_unit_prefix = '%s_%s' % (io_type_prefix, lat_unit)
                         for lat_percentile in ['95.000000', '99.000000', '99.900000']:
                             lat = fio_output['jobs'][0][io_type][lat_unit]['percentile'][lat_percentile]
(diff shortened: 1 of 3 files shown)


abuccts (approved):

tests/benchmarks/micro_benchmarks/test_disk_performance.py
```diff
@@ -504,34 +504,34 @@ def test_disk_performance_result_parsing(self):
   ]
 }
 """
-        jobname_prefix = 'disk_performance:/dev/nvme0n1:rand_read_write'
+        jobname_prefix = '/dev/nvme0n1_rand_read_write'
```
guoshzhao, line 507: Do we have `/dev/` in the metric name? If so, could we remove `/dev/`?

yzygitzh: Fixed.

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -190,25 +190,25 @@ def _process_raw_result(self, cmd_idx, raw_output):
             fio_output = json.loads(raw_output)
 
             jobname = fio_output['jobs'][0]['jobname']
```
cp5555, line 192: Is it a bug? It's like a revision for the result summary.

yzygitzh: It is, because current metric name uses reserved character ':', which might cause metric post processing failure.

cp5555: Yes. Therefore, it's not a bug. It's the revision to support metric post processing.

yzygitzh: Shall we split this into 2 PRs?

cp5555: You can revise your title directly.

yzygitzh: Fixed.

guoshzhao (approved):

PR #176, opened 2021-08-30
Benchmarks: Add Configuration - Add microbenchmark in the validation config file for HPE (AMD MI00)

yukirora:
**Description**
Add microbenchmark in the validation config file for AMD MI00.

**Major Revision**
- add rccl-bw, mem-bw,ib-loopback,gemm-flops,kernel-launch config for mi100


superbench/config/amd_mi100_z53.yaml
```diff
@@ -36,6 +36,39 @@ superbench:
         - train
       pin_memory: yes
   benchmarks:
+    kernel-launch:
+      <<: *default_local_mode
+    rccl-bw:
+      enable: true
+      modes:
+        - name: local
+          prefix: NCCL_DEBUG=INFO NCCL_IB_DISABLE=1
```
abuccts, line 45: why INFO

superbench/config/amd_mi100_z53.yaml
```diff
@@ -36,6 +36,39 @@ superbench:
...
+      <<: *default_local_mode
+      parameters:
+        m: 7680 
+        n: 8192 
+        k: 8192
+    ib-loopback:
+      enable: true
+      modes:
+        - name: local
+          prefix: PROC_RANK={proc_rank} IB_DEVICES=0
```
abuccts, line 61: do you have proc_num?

superbench/config/amd_mi100_z53.yaml
```diff
@@ -36,6 +36,39 @@ superbench:
...
+          parallel: no
+        - name: local
```
abuccts, line 63: we do not support multiple modes currently because cannot diff metrics

yukirora: the metric in ib-loopback will be diffed by ib index, it works fine.

abuccts: if you only run 4 processes, why not `PROC_RANK={proc_rank} IB_DEVICES={proc_rank}` directly?

yukirora: it will affect the results,they'll use same numa cores

superbench/config/amd_mi100_z53.yaml
yukirora: removed

yzygitzh:
For disk benchmark, I think you can check the list of NMVe devices in Z53 and add it accordingly. For GPU SM copy benchmark, I think you can check number of NUMA nodes and GPUs and add it accordingly. I'm not sure whether we should enable these two tests by default, because they would take relatively long time.

superbench/config/amd_mi100_z53.yaml
guoshzhao: Use？
```
proc_num: 4
PROC_RANK={proc_rank} IB_DEVICES=0,1,2,3
parallel: no
```

guoshzhao (approved):

abuccts (approved):

superbench/config/amd_mi100_hpe.yaml
```diff
@@ -0,0 +1,140 @@
+# SuperBench Config
+#
+# Server:
+#   - Product: HPE Apollo 6500
+
+version: v0.3
```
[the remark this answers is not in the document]
yukirora: updated

PR #178, opened 2021-08-30
Benchmarks: Code Revision - Revise subprocess invoke

guoshzhao:
**Description**
Package frequently-used subprocess invoke into function.

### DIFF
--- a/superbench/common/utils/process.py
+++ b/superbench/common/utils/process.py
@@ -1,0 +1,22 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Process Utility."""
+
+import subprocess
+
+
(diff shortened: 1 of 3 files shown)


superbench/common/utils/process.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) Microsoft Corporation.
```
cp5555, line 1: The code logic looks like the same, correct?

cp5555 (approved):

superbench/common/utils/process.py
guoshzhao: Yes, code logic remain unchanged.

abuccts (approved):

PR #185, opened 2021-09-01
Runner - Fix inventory issue in ansible_runner

abuccts:
__Description__

Fix inventory bug in ansible_runner when host list is provided with multiple hosts.

It ought to be handled by ansible_runner lib, workaround by using `--inventory` arg in cmdline.


superbench/runner/ansible.py
```diff
@@ -33,14 +32,14 @@ def __init__(self, config):
...
             if inventory_file or inventory_list:
-                self._config['inventory'] = inventory_file or inventory_list
                 self._config['host_pattern'] = 'all'
                 inventory = InventoryManager(loader=DataLoader(), sources=inventory_file or f'{inventory_list},')
                 host_list = inventory.get_groups_dict()['all']
                 if len(host_list) > 0:
                     self._config['cmdline'] = '--forks {}'.format(len(host_list))
                 if inventory_list in ['localhost', '127.0.0.1']:
                     self._config['cmdline'] += ' --connection local'
+                self._config['cmdline'] += ' --inventory {}'.format(inventory_file or f'{inventory_list},')
```
guoshzhao, line 42: Why add `,` at the end?

guoshzhao (approved):

superbench/runner/ansible.py
cp5555: if you do not add `, ` , it will be treated as a file name. Reference: https://stackoverflow.com/a/48155344/7630831

PR #194, opened 2021-09-09
CI/CD - Add ROCm image build in GitHub Actions

abuccts:
Add ROCm image build in GitHub Actions.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -10,9 +10,18 @@ on:
     - release/*
 
 jobs:
-  cuda-image:
-    name: Build CUDA image
+  docker:
+    name: Docker build
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+        - tags: superbench/superbench:dev-cuda11.1.1,superbench/superbench:latest
+          dockerfile: cuda11.1.1.dockerfile
+        - tags: superbench/superbench:dev-rocm4.2
+          dockerfile: rocm4.2-pytorch1.7.0.dockerfile
+        - tags: superbench/superbench:dev-rocm4.0
+          dockerfile: rocm4.0-pytorch1.7.0.dockerfile
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


cp5555 (approved):

.github/workflows/build-image.yml
```diff
@@ -72,3 +72,65 @@ jobs:
             org.opencontainers.image.revision=${{ github.sha }}
       - name: Echo image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
+  rocm-image:
```
cp5555, line 75: It will build the image in sequential. Correct?

.github/workflows/build-image.yml
```diff
@@ -72,3 +72,65 @@ jobs:
...
+            sudo rsync -a --delete /tmp/emptydir/ ${dir}
+          done
+          sudo apt-get clean
+          df -h
+          echo 'nproc: '$(nproc)
+      - name: Prepare metadata
+        id: metadata
+        run: |
+          DOCKER_IMAGE=superbench/superbench
+          IMAGE_TAG=dev-rocm4.2
```
guoshzhao, line 96: Use `dev-rocm4.2-pytorch1.7.0`? AMD will release torch1.7 and torch1.8 images.

.github/workflows/build-image.yml
abuccts: no, it will maximize the number of jobs run in parallel depending on the available vms

.github/workflows/build-image.yml
abuccts: will we support both at the same time?

guoshzhao: Yes, we may run both in the future.

abuccts: what if there's pytorch 1.7, 1.8, 1.9, 1.10, ... for each rocm version?
I don't think we can support so many rocm images and create so many criteria for every version combination.

guoshzhao: Not exactly. We don't need to support so many ROCm images, for ROCm4.2 and 4.3, we just need to support 1.7 and 1.8.
For this case, how to you distinguish the image?

abuccts: > we just need to support 1.7 and 1.8
why do you choose 1.7 and 1.8 but do not choose 1.9?

guoshzhao: I think we have discussed in previous scrum? Copied from our slides:
```
SuperBench v0.2 AMD docker consolidation
ROCm 4.0 / 4.2
PyTorch 1.7.0 / 1.8.0
```

guoshzhao: No matter what pytorch version to provide, we should have the mechanism to distinguish images with same driver and different pytorch which will affect the performance a lot. They are different problem:
1. what pytorch version to provide, it's decided by us
2. mechanism to distinguish images.

We should firstly support 2, then to handle 1 according to the requirement.

For your change, we can only provide one image for specific ROCm version.

guoshzhao (approved):

cp5555 (approved):

PR #198, opened 2021-09-10
Bug: Fix Bug - Add barrier before 'destroy_process_group' in model benchmarks

yukirora:
**Description**
Add barrier before 'destroy_process_group' to resolve the bug due to when multi models in one model benchmark, some processes haven't finished the previous process group while others failed to initialize new process group for the next model on rocm4.x when running bert_models.

**Major Revision**
-  Add barrier before 'destroy_process_group'.


abuccts (commented):
to workaround, could you update the config file for amd?
otherwise, the model results will override previous ones in the same benchmark

abuccts (approved):

cp5555 (approved):

guoshzhao (approved):

PR #199, opened 2021-09-14
CLI - Integrate system info for node

abuccts:
Integrate system info for node, add `sb node info` command.

### DIFF
--- a/superbench/cli/_commands.py
+++ b/superbench/cli/_commands.py
@@ -23,6 +23,8 @@ class SuperBenchCommandsLoader(CLICommandsLoader):
             g.command('deploy', 'deploy_command_handler')
             g.command('exec', 'exec_command_handler')
             g.command('run', 'run_command_handler')
+        with CommandGroup(self, 'node', 'superbench.cli._node_handler#{}') as g:
+            g.command('info', 'info_command_handler')
         return super().load_command_table(args)
 
     def load_arguments(self, command):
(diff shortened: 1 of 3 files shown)


superbench/cli/_commands.py
```diff
@@ -23,6 +23,8 @@ def load_command_table(self, args):
             g.command('deploy', 'deploy_command_handler')
             g.command('exec', 'exec_command_handler')
             g.command('run', 'run_command_handler')
+        with CommandGroup(self, 'node', 'superbench.cli._node_handler#{}') as g:
```
cp5555, line 26: What's the meaning of this?

superbench/cli/_commands.py
```diff
@@ -23,6 +23,8 @@ def load_command_table(self, args):
...
+            g.command('info', 'info_command_handler')
```
cp5555, line 27: How can we collect all information from all nodes?

superbench/cli/_commands.py
abuccts: create a command group https://github.com/microsoft/knack/blob/dev/docs/commands.md#commands

superbench/cli/_commands.py
abuccts: it can only be used for one node
the cli is even not installed on any compute node

cp5555: Shall we add an ansible task for this? Maybe we can do it in future. I just think we can leverage this to check the configuration.

abuccts: it's not aligned with current design (cli is only run in container, host does not install cli, but the collection needs to be run on host)
we can add such a feature later

cp5555 (approved):

yukirora (approved):

PR #217, opened 2021-09-29
Benchmarks: Add Benchmark - Add tcp connectivity validation microbenchmark

yukirora:
**Description**
Add tcp connectivity validation microbenchmark which is to validate TCP connectivity between current node and several nodes in the hostfile.

**Major Revision**
- Add tcp connectivity validation microbenchmark and related test, example

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/tcp_connectivity.py
+++ b/superbench/benchmarks/micro_benchmarks/tcp_connectivity.py
@@ -1,0 +1,204 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""Module of the TCP connectivity benchmarks."""
+
+import socket
+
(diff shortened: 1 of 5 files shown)


superbench/benchmarks/micro_benchmarks/tcp_connectivity.py
```diff
@@ -0,0 +1,204 @@
...
+            '--hostfile',
+            type=str,
+            default='/root/hostfile',
+            required=False,
+            help='The path of the hostfile including addresses to test',
+        )
+        self._parser.add_argument(
+            '--port',
+            type=int,
+            default=22222,
```
cp5555, line 66: @abuccts do you think we need to change to typical SSH port?

cp5555 (approved):

PR #239, opened 2021-11-11
Benchmarks: Add Feature - Extend the device manager utility to support more functions.

guoshzhao:
**Description**
Rename `nvidia_helper` utility as `device_manager` module and support more functions:
```
device_manager.get_device_count()
device_manager.get_device_utilization(idx)
device_manager.get_device_temperature(idx)
device_manager.get_device_power_limit(idx)
device_manager.get_device_memory(idx)
device_manager.get_device_row_remapped_info(idx)
device_manager.get_device_ecc_error(idx)
```


superbench/common/utils/nvidia_helper.py
```diff
@@ -5,18 +5,180 @@
...
+        dbecc (int): the count of double bit ecc error, None means failed to get the data.
+    """
+    output = run_command('nvidia-smi dmon -i {} -s e -c 1'.format(idx))
+    if output.returncode == 0:
+        content = output.stdout.splitlines()
+        if len(content) == 3:
+            ecc = list(filter(None, content[2].split(' ')))
+            if len(ecc) == 4:
+                sbecc = int(ecc[1])
+                dbecc = int(ecc[2])
```
abuccts, line 181: what if they are 'N/A' on geforce cards?

superbench/common/utils/nvidia_helper.py
```diff
@@ -5,18 +5,180 @@
...
+        Temperature
+            GPU Current Temp                  : 36 C
+
+    Args:
+        idx (int): device index.
+
+    Return:
+        remapped_metrics (dict): the row remapped information, None means failed to get the data.
+    """
+    output = run_command('nvidia-smi -i {} -q'.format(idx))
```
abuccts, line 135: why not use `nvmlDeviceGetRemappedRows`?

superbench/common/utils/nvidia_helper.py
abuccts: use `nvmlDeviceGetDetailedEccErrors`?

superbench/common/utils/nvidia_helper.py
guoshzhao: `nvmlDeviceGetRemappedRows` doesn't have python API support.

superbench/common/utils/nvidia_helper.py
guoshzhao: `nvmlDeviceGetDetailedEccErrors` is labelled as deprecated:
https://github.com/fbcotter/py3nvml/blob/0aa41c0940c7d971381fddb8382266a819be9f75/py3nvml/py3nvml.py#L3437

> 
>      * @deprecated   This API supports only a fixed set of ECC error locations
>      *               On different GPU architectures different locations are supported
>      *               See \ref nvmlDeviceGetMemoryErrorCounter

I will use `nvmlDeviceGetMemoryErrorCounter` instead, which use correctable for single bit and uncorrectable for double bit.

superbench/common/utils/nvidia_helper.py
```diff
@@ -5,18 +5,185 @@
...
+    corrected_ecc = 0
+    uncorrected_ecc = 0
+    for location_idx in range(nvml.NVML_MEMORY_LOCATION_COUNT):
+        try:
+            count = nvml.nvmlDeviceGetMemoryErrorCounter(
+                handle, nvml.NVML_MEMORY_ERROR_TYPE_CORRECTED, nvml.NVML_VOLATILE_ECC, location_idx
+            )
+            corrected_ecc += count
+        except nvml.NVMLError:
+            pass
```
cp5555, line 179: If it doesn't work, we cannot know it.

superbench/common/utils/nvidia_helper.py
```diff
@@ -5,18 +5,185 @@
...
+
+        try:
+            count = nvml.nvmlDeviceGetMemoryErrorCounter(
+                handle, nvml.NVML_MEMORY_ERROR_TYPE_UNCORRECTED, nvml.NVML_VOLATILE_ECC, location_idx
+            )
+            uncorrected_ecc += count
+        except nvml.NVMLError:
+            pass
```
cp5555, line 187: The same question for the `pass`. It look like we need to know the error.

superbench/common/utils/nvidia_helper.py
cp5555, line 179: If it doesn't work, we cannot know it.

cp5555: Shall we implement it as get_device_memory?

superbench/common/utils/nvidia_helper.py
```diff
@@ -5,18 +5,185 @@
 
 import py3nvml.py3nvml as nvml
 
+from superbench.common.utils import logger
+from superbench.common.utils import run_command
+
+nvml.nvmlInit()
```
cp5555, line 11: It's a little bit weird to init it globally.

cp5555: It's quite like we should provide a class on this. It can init nvml, handle , and util.

cp5555: We can use this class as logger.

superbench/common/utils/nvidia_helper.py
guoshzhao: There have 8 types of memory location, for the nvml API design, the ecc counter is the total ecc error occurred in all the location:
```
class nvmlMemoryLocation_t(Enum):
    NVML_MEMORY_LOCATION_L1_CACHE = 0
    NVML_MEMORY_LOCATION_L2_CACHE = 1
    NVML_MEMORY_LOCATION_DRAM = 2
    NVML_MEMORY_LOCATION_DEVICE_MEMORY = 2
    NVML_MEMORY_LOCATION_REGISTER_FILE = 3
    NVML_MEMORY_LOCATION_TEXTURE_MEMORY = 4
    NVML_MEMORY_LOCATION_TEXTURE_SHM = 5
    NVML_MEMORY_LOCATION_CBU = 6
    NVML_MEMORY_LOCATION_SRAM = 7
    NVML_MEMORY_LOCATION_COUNT = 8
```
Different arch supports different locations, for example, A100 has `NVML_MEMORY_LOCATION_SRAM` supported but p100 doesn't. P100 has `NVML_MEMORY_LOCATION_TEXTURE_SHM` supported but A100 doesn't.

The error raised here is `NVMLError_NotSupported` for current device. The current implementation can ignore the device type and loop all the location to aggregate the ecc error.

Your suggestion is naming it as `get_device_memory`? Do you mean we can only get the ecc count from `NVML_MEMORY_LOCATION_DEVICE_MEMORY` ?

guoshzhao: Currently nvml.NVMLError uses `pass` and other type of exceptions will log error.

superbench/common/utils/nvidia_helper.py
guoshzhao: Done, and abstract the utility as a class.

cp5555 (approved):

abuccts (approved):

PR #240, opened 2021-11-15
Monitor: Initialization - Add Monitor and MonitorRecord class

guoshzhao:
**Description**
Add the initial version of Monitor.

**Major Revision**
- Add `Monitor` class to launch background process for monitoring.
- Add `MonitorRecord` class to save the data one time capturing.


superbench/monitor/monitor.py
```diff
@@ -0,0 +1,261 @@
...
+import glob
+import sched
+import multiprocessing
+
+from superbench.common.utils import logger, run_command
+from superbench.common.utils import device_manager as dm
+from superbench.monitor.record import MonitorRecord
+
+
+class Monitor(multiprocessing.Process):
```
cp5555, line 17: Just one general comment: We don't have child class of Monitor, therefore, it may change to __ for variable and function

cp5555 (approved):

superbench/monitor/monitor.py
```diff
@@ -0,0 +1,268 @@
...
+        record.net_receive = net_receive
+        record.net_transmit = net_transmit
+
+    def __sample_gpu_metrics(self, record):
+        """Method sampling the gpu metrics.
+
+        Args:
+            record (MonitorRecord): record instance to save the metrics.
+        """
+        gpu_usage = list()
```
abuccts, line 174: use `record.gpu_usage` directly? why need a temporal variable?

abuccts (approved):

superbench/monitor/monitor.py
guoshzhao: Done.

PR #245, opened 2021-11-24
Benchmarks: Add Benchmark - Add ONNXRuntime inference benchmark based on ORT python API

guoshzhao:
**Description**
Add ONNXRuntime inference benchmark based on ORT python API.

**Major Revision**
- Add `ORTInferenceBenchmark` class to export pytorch model to onnx model and do inference
- Add tests and example for `ort-inference` benchmark
- Update the introduction docs.


docs/user-tutorial/benchmarks/micro-benchmarks.md
```diff
@@ -83,6 +83,18 @@ Inference PyTorch/ONNX models on NVIDIA GPUs with [TensorRT](https://developer.n
 | tensorrt-inference/end_to_end_lat_ms_mean | time (ms) | The mean duration from when the H2D of a query is called to when the D2H of the same query is completed. |
 | tensorrt-inference/end_to_end_lat_ms_99   | time (ms) | The P99 duration from when the H2D of a query is called to when the D2H of the same query is completed.  |
 
+### `ort-inference`
+
+#### Introduction
+
+Inference performance of the torchvision models using ONNXRuntime.
```
cp5555, line 90: Would you please add the support model list here?

examples/benchmarks/ort_inference_performance.py
```diff
@@ -0,0 +1,23 @@
...
+    context = BenchmarkRegistry.create_benchmark_context(
+        'ort-inference', platform=Platform.CUDA, parameters='--pytorch_models resnet50 resnet101 --precision float16'
+    )
+    benchmark = BenchmarkRegistry.launch_benchmark(context)
+    if benchmark:
+        logger.info(
+            'benchmark: {}, return code: {}, result: {}'.format(
+                benchmark.name, benchmark.return_code, benchmark.result
+            )
+        )
```
cp5555, line 23: Does this ort support AMD as well?

dockerfile/cuda11.1.1.dockerfile
```diff
@@ -108,5 +108,5 @@ ADD third_party third_party
 RUN make -j -C third_party cuda
 
 ADD . .
-RUN python3 -m pip install .[nvidia,torch] && \
+RUN python3 -m pip install .[nvidia,torch,ort] && \
```
cp5555, line 111: Just check. Do we need to consider AMD?

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
```diff
@@ -0,0 +1,187 @@
...
+
+import time
+import statistics
+from pathlib import Path
+
+import torch
+import torchvision.models
+import numpy as np
+import onnxruntime as ort
+from onnxruntime.quantization import quantize_dynamic
```
cp5555, line 14: We already have had onnxruntime. Why not use ort.quantization directly because we just use once.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
```diff
@@ -0,0 +1,187 @@
...
+            'resnet50',
+            'resnet101',
+            'resnet152',
+            'densenet169',
+            'densenet201',
+            'vgg11',
+            'vgg13',
+            'vgg16',
+            'vgg19',
+        ]
```
cp5555, line 42: Do we have bert/gpt/lstm ? Actually I have the same question on Yifan's PR as well.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
```diff
@@ -0,0 +1,187 @@
...
+            choices=[0, 1, 2, 3],
+            required=False,
+            help='ONNXRuntime graph optimization level, 0 for ORT_DISABLE_ALL, 1 for ORT_ENABLE_BASIC, '
+            '2 for ORT_ENABLE_EXTENDED, 3 for ORT_ENABLE_ALL.',
+        )
+
+        self._parser.add_argument(
+            '--batch_size',
+            type=int,
+            default=32,
```
cp5555, line 85: Do we need to change the default batch size to 1? It's also for Yifan's TensorRT PR

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
```diff
@@ -0,0 +1,187 @@
...
+                data_type = Precision.FLOAT16.value if self._args.precision == Precision.FLOAT16 \
+                    else Precision.FLOAT32.value
+                model_path = f'{self.__model_cache_path / (model + "." + data_type + ".onnx")}'
+                torch.onnx.export(
+                    getattr(torchvision.models, model)(pretrained=True).to(dtype=getattr(torch, data_type)).cuda(),
+                    torch.randn(self._args.batch_size, 3, 224, 224, device='cuda', dtype=getattr(torch, data_type)),
+                    model_path,
+                    input_names=['input'],
+                )
+                if self._args.precision == Precision.INT8:
```
cp5555, line 126: Why only INT8 need quantize?

docs/user-tutorial/benchmarks/micro-benchmarks.md
guoshzhao: Done. Same with our `pytorch_cnn.py` benchmarks.

examples/benchmarks/ort_inference_performance.py
guoshzhao: It can run on AMD platform. But I found that on AMD, the inference happens on CPU not GPU even the models are saved on GPU device. This doesn't happen on CUDA platform.
I think we can look into it in the future.

dockerfile/cuda11.1.1.dockerfile
guoshzhao: There have no difference between CUDA and AMD for package installation.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
guoshzhao: Done.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
guoshzhao: I ever thought you guys have reached the agreement about what models to support in Yifan's PR.....

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
guoshzhao: For all the default values of benchmarks, we used the most common values, not the least value.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
guoshzhao: Quantization in ONNX Runtime refers to 8 bit linear quantization of an ONNX model. Please refer:
https://onnxruntime.ai/docs/performance/quantization.html#quantization-overview

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
guoshzhao: For other models like GPT, the model file size will be big. We can discuss what models to support in the future.

examples/benchmarks/ort_inference_performance.py
cp5555: Thanks!

docs/user-tutorial/benchmarks/micro-benchmarks.md
cp5555: Thanks

dockerfile/cuda11.1.1.dockerfile
cp5555: Got it.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
cp5555: Got it. Make sense!

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
cp5555: Got it.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
cp5555: Shall we add a comment here including this link?

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
cp5555: Do you think 32 is reasonable for inference? It's my question here.

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
```diff
@@ -0,0 +1,187 @@
...
+                    file_name = '{model}.{precision}.onnx'.format(model=model, precision=self._args.precision)
+                    quantize_dynamic(model_path, f'{self.__model_cache_path / file_name}')
```
cp5555, line 128: Besides, based on the document,
```
In general, it is recommended to use dynamic quantization for RNN and transformer-based models, and static quantization for CNN models.
```

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
guoshzhao: There is a balance issue among latency/throughput/resource utilization when using batchsize as 1 and greater than 1. For inference by using GPU, I think batch size greater than 1 is reasonable, since GPU resource is not shared with other services, batch-size as 1 is very wasteful and I think no one will use in this way. For inference by CPU, we can set it as 1 to get the smallest latency.
Here we use the benchmarks for GPU validation, so we can set the default value as 32 like model training benchmarks. Do you think it is reasonable?

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
guoshzhao: As the doc introduced,
> The main difference between dynamic quantization and static quantization is how scale and zero point of activation is calculated. For static quantization, they are calculated offline with calibration data set. All the activations have same scale and zero point. While for dynamic quantization, they are calculated on flight and will be specific for each activation, thus they are more accurate but introduce extra computation overhead.

So static quantization can be regarded as an improvement feature, which improved the performance but lost some accuracy compared with dynamic quantization. For CNN models, the memory bandwidth for activations is more heavy than transformer-based models, so the docs suggest to use static quantization for CNN models.
I suggest we use the dynamic quantization which is a more common way. For static quantization, we have to prepare one calibration dataset for offline calculation.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
guoshzhao: Added.

docs/user-tutorial/benchmarks/micro-benchmarks.md
```diff
@@ -71,6 +71,12 @@ TODO
 #### Introduction
 
 Inference PyTorch/ONNX models on NVIDIA GPUs with [TensorRT](https://developer.nvidia.com/tensorrt).
+Currently the following models are supported:
+
+> 'alexnet', 'densenet121', 'densenet169', 'densenet201', 'densenet161', 'googlenet', 'inception_v3', 'mnasnet0_5',
```
abuccts, line 76: do not need quotes?

tests/benchmarks/micro_benchmarks/test_ort_inference_performance.py
```diff
@@ -0,0 +1,63 @@
...
+from tests.helper import decorator
+from superbench.benchmarks import BenchmarkRegistry, Platform, Precision, BenchmarkType, ReturnCode
+from superbench.benchmarks.micro_benchmarks.ort_inference_performance import ORTInferenceBenchmark
+
+
+@decorator.cuda_test
+@decorator.pytorch_test
+@mock.patch('torch.hub.get_dir')
+@mock.patch('onnxruntime.InferenceSession.run')
+def test_ort_inference_performance(mock_ort_session_run, mock_get_dir):
```
abuccts, line 22: is `mock_ort_session_run` actually used?

tests/benchmarks/micro_benchmarks/test_ort_inference_performance.py
```diff
@@ -0,0 +1,63 @@
...
+
+from tests.helper import decorator
+from superbench.benchmarks import BenchmarkRegistry, Platform, Precision, BenchmarkType, ReturnCode
+from superbench.benchmarks.micro_benchmarks.ort_inference_performance import ORTInferenceBenchmark
+
+
+@decorator.cuda_test
+@decorator.pytorch_test
+@mock.patch('torch.hub.get_dir')
+@mock.patch('onnxruntime.InferenceSession.run')
```
abuccts, line 21: will the test actually run the inference?

abuccts (approved):

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
cp5555: Shall we check it on MLPerf? Whether it's a large number?

tests/benchmarks/micro_benchmarks/test_ort_inference_performance.py
guoshzhao: No, the `InferenceSession.run` will be mocked and it just records the inference latency as very small value.

tests/benchmarks/micro_benchmarks/test_ort_inference_performance.py
guoshzhao: Yes, it is used.

superbench/benchmarks/micro_benchmarks/ort_inference_performance.py
cp5555: Yifan said that Nvidia uses 32.

PR #247, opened 2021-11-29
Benchmarks: Add Feature - Add 'ignore_invalid' option when register benchmarks.

guoshzhao:
**Description**
If `ignore_invalid` is True, and 'required' arguments are not set when register the benchmark, the arguments should be provided by user in config and skip the arguments checking.

### DIFF
--- a/superbench/benchmarks/base.py
+++ b/superbench/benchmarks/base.py
@@ -60,7 +60,7 @@ class Benchmark(ABC):
         """
         return self._parser.format_help().strip()
 
-    def parse_args(self):
+    def parse_args(self, ignore_required=False):
         """Parse the arguments.
 
         Return:
@@ -71,8 +71,12 @@ class Benchmark(ABC):
         try:
             args, unknown = self._parser.parse_known_args(self._argv)
         except BaseException as e:
-            logger.error('Invalid argument - benchmark: {}, message: {}.'.format(self._name, str(e)))
-            return False, None, None
+            if ignore_required and isinstance(e, SystemExit):
+                logger.info('Missing required parameter, will ignore the error and skip the args checking.')
+                return True, None, []
+            else:
+                logger.error('Invalid argument - benchmark: {}, message: {}.'.format(self._name, str(e)))
+                return False, None, []
 
         ret = True
         if len(unknown) > 0:
--- a/superbench/benchmarks/base.py
+++ b/superbench/benchmarks/base.py
@@ -60,7 +60,7 @@ class Benchmark(ABC):
         """
         return self._parser.format_help().strip()
 
-    def parse_args(self, ignore_required=False):
+    def parse_args(self, ignore_invalid=False):
         """Parse the arguments.
 
         Return:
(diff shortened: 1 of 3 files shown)


superbench/benchmarks/base.py
```diff
@@ -71,8 +71,12 @@ def parse_args(self):
         try:
             args, unknown = self._parser.parse_known_args(self._argv)
         except BaseException as e:
-            logger.error('Invalid argument - benchmark: {}, message: {}.'.format(self._name, str(e)))
-            return False, None, None
+            if ignore_required and isinstance(e, SystemExit):
```
yzygitzh, line 74: Looks like we cannot tell the difference among 3 cases: `wrong argument specified`, `required argument not given` and the combination of both.
Maybe a better way could be letting benchmark specify required parameters in an additional `required_parameters` dict, as a parameter of `register_benchmark`. `register_benchmark` can then use that dict with existing `parameters` to pass `parse_args`'s check, but still using `parameters` to fill predefined_args.
What do you think?

yzygitzh (changes requested):

superbench/benchmarks/base.py
guoshzhao: We can checking the wrong cases when run benchmarks, not during the registration. Just like lazy checking.

guoshzhao: When run the benchmark, sb will call the `parse_args` with `ignore_required=False`, which can cover all the wrong cases.

yzygitzh: Actually I mean whether `SystemExit` exception is the difference between `wrong argument specified` and `required argument not given`? If yes I think it's ok.

guoshzhao: There has one option `exit_on_error` for `argparse` module. If False, only `required argument not given` raise the `SystemExit` exception. But the default value is True, means all wrong case will raise the `SystemExit` exception:
https://github.com/python/cpython/blob/3.10/Lib/argparse.py#L2091
https://github.com/python/cpython/blob/3.10/Lib/argparse.py#L1852

What I mean is that, the argument checking is not must when do benchmark registration, if the checking can be done earlier, we can just do it. Otherwise, the checking will be done again when launch the benchmark. Superbench can handle all the wrong cases and give the error logs.

yzygitzh: Get your point about delaying the parameter checking for `required param` case. But what if a benchmark has `predefine_param` and `required param` at the same time?

guoshzhao: Looks `exit_on_error` is introduced with python>=3.9. I have updated the log description to combine the `wrong argument specified` and `required argument not given` error, we don't need to distinguish them when `ignore_invalid=True`, could you please check again?

yzygitzh: Just checked that, please check my latest comment. Thanks!

guoshzhao: In current implementation, user can set `required params` in `predefine params`  or set them in config files. If errors happen when register the benchmark, we can ignore the errors and check combined params (predefine params + params in config file) again when launch the benchmark.
For the params checking/parsing when registration, it is mainly for the function `get_all_benchmark_predefine_settings`, which is an interface to get all the predefine/default params of benchmarks.

yzygitzh (approved):

tests/benchmarks/test_registry.py
```diff
@@ -28,8 +28,7 @@ def add_parser_arguments(self):
         self._parser.add_argument(
             '--lower_bound',
             type=int,
-            default=0,
-            required=False,
+            required=True,
```
cp5555, line 31: Do we need to add a test for this? It's a little wired that we revise the test.

cp5555 (approved):

abuccts (approved):

tests/benchmarks/test_registry.py
guoshzhao: The current test case can trigger and test it with this change.

PR #261, opened 2021-12-10
Benchmarks: Fix Bug - Set reduce_op type for metirc return_code

guoshzhao:
**Description**
Set the `reduce_op` type for metirc `return_code` as `None`.


cp5555 (approved):

superbench/benchmarks/result.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, name, type, return_code, run_count=0):
         self.__result = dict()
         self.__result['return_code'] = [return_code.value]
         self.__reduce_op = dict()
+        self.__reduce_op['return_code'] = None
```
abuccts, line 36: why it is None, not Last etc.?

guoshzhao: This is metric `return_code`, every worker should have its own return_code, don't need to do reduce.

abuccts (approved):

superbench/benchmarks/result.py
abuccts: because it's a list, is it possible to have multiple return codes? e.g., benchmark itself add a metric with same name

guoshzhao: Here the list is for `run_count`. It is aligned with other metrics, but for `return_code`, we only have one for each BenchmarkResult instance. I think Benchmark itself shouldn't have metric named as `return_code`.

PR #267, opened 2021-12-15
Bug - Fix issues for Ansible and benchmarks

abuccts:
__Description__

Fix issues for Ansible and benchmarks:
* Cleanup Ansible runner private data dir to avoid out of disk space issue when node number is large.
* Support both absolute and relative paths when fecth results.
* Use a deterministic image in Ansible test to avoid image update.
* Update logging format.
* Delete torch models and inputs after export.

### DIFF
--- a/tests/ansible/tests/test_deploy.yaml
+++ b/tests/ansible/tests/test_deploy.yaml
@@ -14,4 +14,4 @@
   vars:
     ssh_port: 12345
     output_dir: /tmp/test_ansible
-    docker_image: superbench/superbench
+    docker_image: superbench/superbench:v0.3.0-cuda11.1.1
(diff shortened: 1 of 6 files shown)


tests/ansible/tests/test_deploy.yaml
```diff
@@ -14,4 +14,4 @@
   vars:
     ssh_port: 12345
     output_dir: /tmp/test_ansible
-    docker_image: superbench/superbench
+    docker_image: superbench/superbench:v0.3.0-cuda11.1.1
```
cp5555, line 17: Do we need to change this?

abuccts: to avoid frequent image update and no space issue

yzygitzh (approved):

tests/ansible/tests/test_deploy.yaml
abuccts: add comment, the image requires `sb` installed

cp5555: Got it, thanks!

cp5555 (approved):

PR #270, opened 2021-12-20
Bug - Fix issue that the root mpi rank may not be the first in the hostfile

yukirora:
**Description**
Launch mpi on the sorted first host in the hostfile.


tests/runner/test_ansible.py
```diff
@@ -22,8 +22,8 @@ def setUp(self):
             fd, (
                 'all:\n'
                 '  hosts:\n'
-                '    10.0.0.10:\n'
                 '    10.0.0.11:\n'
+                '    10.0.0.10:\n'
```
cp5555, line 26: Why not add a new test?

superbench/runner/ansible.py
```diff
@@ -34,9 +34,9 @@ def __init__(self, config):
             if inventory_file or inventory_list:
                 self._config['host_pattern'] = 'all'
                 inventory = InventoryManager(loader=DataLoader(), sources=inventory_file or f'{inventory_list},')
-                host_list = inventory.get_groups_dict()['all']
-                if len(host_list) > 0:
-                    self._config['cmdline'] = '--forks {}'.format(len(host_list))
+                self._host_list = inventory.get_hosts(pattern='all', order='sorted')
```
cp5555, line 37: Why `self._host_list`? Do we need to store it?

yukirora: we need to at least store the sorted first host in this init() method, which will be used in update_mpi_config() method.

tests/runner/test_ansible.py
yukirora: added

cp5555 (approved):

superbench/runner/ansible.py
```diff
@@ -34,7 +34,8 @@ def __init__(self, config):
             if inventory_file or inventory_list:
                 self._config['host_pattern'] = 'all'
                 inventory = InventoryManager(loader=DataLoader(), sources=inventory_file or f'{inventory_list},')
-                host_list = inventory.get_groups_dict()['all']
+                host_list = inventory.get_hosts(pattern='all', order='sorted')
+                self._first_host = host_list[0].get_name()
                 if len(host_list) > 0:
```
abuccts, line 39: rename to head_host?

abuccts: what if the length is 0?

tests/runner/test_ansible.py
```diff
@@ -61,7 +61,34 @@ def test_update_mpi_config(self):
         self.assertDictEqual(
             self.ansible_client.update_mpi_config(self.ansible_client._config), {
                 **self.ansible_client._config,
-                'host_pattern': 'all[0]',
+                'host_pattern': '10.0.0.10',
+            }
+        )
+        # Test out-of-order hosts defined in inventory
+        fd = os.open(self.host_file, os.O_RDWR)
+        os.write(
```
abuccts, line 69: could you write the original file instead of changing it in the middle? what if other test cases need to test the original file?

abuccts (changes requested):

tests/runner/test_ansible.py
yukirora: changed to use another file

superbench/runner/ansible.py
yukirora: updated, it will raise exception in update_mpi_config

superbench/runner/ansible.py
```diff
@@ -87,7 +89,9 @@ def update_mpi_config(self, ansible_config):
         Returns:
             dict: Updated Ansible config dict.
         """
-        ansible_config['host_pattern'] += '[0]'
+        if not self._head_host:
+            logger.log_and_raise('no available host to launch mpi')
+        ansible_config['host_pattern'] = self._head_host
```
abuccts, line 94: if you raise error it will exit and won't run following benchmarks? could you just set a default value like "all[0]" instead of None?

tests/runner/test_ansible.py
```diff
@@ -38,10 +38,12 @@ def setUp(self):
                 'host_password': 'pass',
             })
         )
+        fd, self.host_file2 = tempfile.mkstemp()
```
abuccts, line 41: why need a fd here?

abuccts: could you rename it to a meaningful name?

tests/runner/test_ansible.py
```diff
@@ -61,9 +63,53 @@ def test_update_mpi_config(self):
         self.assertDictEqual(
             self.ansible_client.update_mpi_config(self.ansible_client._config), {
                 **self.ansible_client._config,
-                'host_pattern': 'all[0]',
+                'host_pattern': '10.0.0.10',
             }
         )
+        # Test out-of-order hosts defined in inventory
```
abuccts, line 69: define it in another test case?

tests/runner/test_ansible.py
```diff
@@ -61,9 +63,53 @@ def test_update_mpi_config(self):
...
+        with open(self.host_file2, 'w') as fd:
+            fd.write(
+                'all:\n' + '  hosts:\n' + '    10.0.0.12:\n' + '    10.0.0.11:\n' + '    10.0.0.10:\n' +
```
abuccts, line 72: why need "+" here?

superbench/runner/ansible.py
yukirora: ok, updated

tests/runner/test_ansible.py
yukirora: add a new test function

tests/runner/test_ansible.py
yukirora: update to self.test_mpi_host_file, is it ok?

tests/runner/test_ansible.py
yukirora: deleted

tests/runner/test_ansible.py
```diff
@@ -38,10 +38,12 @@ def setUp(self):
                 'host_password': 'pass',
             })
         )
+        fd, self.test_mpi_host_file = tempfile.mkstemp()
```
abuccts, line 41:
```suggestion
        _, self.test_mpi_host_file = tempfile.mkstemp()
```

abuccts (approved):

PR #287, opened 2022-01-23
Benchmarks: Add Feature - Sync the E2E training results among all workers for each step.

guoshzhao:
**Description**
Please write a brief description and link the related issue if have.

**Major Revision**
- Sync (do allreduce max) the E2E training results among all workers.
- Avoid using ':0' in metric name if there has only one rank having output.

### DIFF
--- a/superbench/benchmarks/model_benchmarks/model_base.py
+++ b/superbench/benchmarks/model_benchmarks/model_base.py
@@ -12,7 +12,6 @@ from superbench.common.utils import logger
 from superbench.benchmarks import Precision, ModelAction, DistributedImpl, DistributedBackend, BenchmarkType, ReturnCode
 from superbench.benchmarks.base import Benchmark
 from superbench.benchmarks.context import Enum
-from superbench.benchmarks.reducer import ReduceType
 
 
 class Optimizer(Enum):
@@ -354,6 +353,17 @@ class ModelBenchmark(Benchmark):
 
         return False
 
+    def _sync_result(self, result):
+        """Function to reduce the result to rank 0.
+
+        Args:
+            result (list): The result data to sync.
+
+        Return:
+            True if reduce result data successfully.
+        """
+        return True
+
     def __process_model_result(self, model_action, precision, step_times):
         """Function to process raw results and save the summarized results.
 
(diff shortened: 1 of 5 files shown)


superbench/benchmarks/model_benchmarks/model_base.py
```diff
@@ -376,22 +395,26 @@ def __process_model_result(self, model_action, precision, step_times):
...
-            self._process_percentile_result(metric, throughput, reduce_type=reduce_type)
+        self._result.add_raw_data(metric_s, step_times)
+        self._result.add_raw_data(metric_t, throughput)
+
+        if model_action == ModelAction.TRAIN:
+            if not self._sync_result(step_times):
+                return False
+            if self._local_rank is None or self._local_rank == 0:
+                self._result.add_result(metric_s, statistics.mean(step_times))
+                throughput = [millisecond_per_second / step_time * self._args.batch_size for step_time in step_times]
```
cp5555, line 411: Why do this? it has been calculated before in line 402

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -172,6 +172,33 @@ def _create_optimizer(self):
...
+            True if reduce result data successfully.
+        """
+        if not super()._sync_result(result):
+            return False
+
+        try:
+            if self._args.distributed_impl == DistributedImpl.DDP:
+                tensor = torch.as_tensor(result)
+                torch.distributed.reduce(tensor, 0, op=torch.distributed.ReduceOp.MAX)
+                result = tensor.tolist()
```
cp5555, line 191: It looks like that the time should be longer than before because we have the Reduce operation one more time.

superbench/runner/runner.py
```diff
@@ -265,6 +265,33 @@ def __create_single_node_summary(self, node_path):    # pragma: no cover # noqa:
 
         return results_summary
 
+    def __generate_metric_name(self, benchmark_name, metric, rank_count, run_count, curr_rank, curr_run):
```
cp5555, line 268: I think we would like to remove the data from other ranks. Why do we need this function.

superbench/benchmarks/model_benchmarks/model_base.py
```diff
@@ -12,7 +12,6 @@
 from superbench.benchmarks import Precision, ModelAction, DistributedImpl, DistributedBackend, BenchmarkType, ReturnCode
 from superbench.benchmarks.base import Benchmark
 from superbench.benchmarks.context import Enum
-from superbench.benchmarks.reducer import ReduceType
```
abuccts, line 15: do we still need the reducer after this PR?

abuccts (commented):
for title, the results are not synced "for each step"? and pls revise the description content

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -172,6 +172,33 @@ def _create_optimizer(self):
...
+        Return:
+            True if reduce result data successfully.
+        """
+        if not super()._sync_result(result):
+            return False
+
+        try:
+            if self._args.distributed_impl == DistributedImpl.DDP:
+                tensor = torch.as_tensor(result)
+                torch.distributed.reduce(tensor, 0, op=torch.distributed.ReduceOp.MAX)
```
abuccts, line 190: did you test this? does it need to create the tensor on gpu?

superbench/runner/runner.py
abuccts: single-gpu test run on multiple gpus? need to differentiate different gpus by ranks in this case

superbench/benchmarks/model_benchmarks/pytorch_base.py
guoshzhao: Yes, but it has no impact on our validation results.

superbench/runner/runner.py
guoshzhao: Correct, this function can work for all benchmarks, take kernel-launch for example, we can have
kernel-launch/overhead_wall:0
kernel-launch/overhead_wall:1
kernel-launch/overhead_wall:2
....

superbench/benchmarks/model_benchmarks/model_base.py
guoshzhao: Reducer can be used for benchmarks which can't be modified to let only rank0 to generate the output.

superbench/benchmarks/model_benchmarks/pytorch_base.py
guoshzhao: Have tested. Since NCCL backend only support collective API on GPU, so we need to transfer the tensor to GPU first.
But for gloo backend, we should use CPU. Maybe I can add a judgement here.

guoshzhao: Looks there have change not pushed...
Have pushed.

superbench/benchmarks/model_benchmarks/model_base.py
cp5555: Why are they different? I think it should be the same.

cp5555 (approved):

superbench/benchmarks/model_benchmarks/model_base.py
guoshzhao: every worker has their own raw data, but only rank0 has the summarized result.

abuccts (approved):

superbench/benchmarks/model_benchmarks/model_base.py
cp5555: Sure. Let's keep it!

PR #288, opened 2022-01-23
Benchmarks: Add Feature - Add timeout feature for each benchmark.

guoshzhao:
**Description**
Add timeout feature for each benchmark.

**Major Revision**
- Add `timeout` config for each benchmark. In current config files, only set the timeout for kernel-launch as example. Other benchmarks can be set in the future.
- Set the timeout config for `ansible_runner.run()`. Runner will get the return code 254:
   [ansible.py:80][WARNING] Run failed, return code 254.
- Using `timeout` command to terminate the client process.


superbench/config/amd_mi100_hpe.yaml
```diff
@@ -37,6 +37,7 @@ superbench:
   benchmarks:
```
cp5555, line 37: Shall we add this feature to document? I think it should be 120 seconds. :)

superbench/config/amd_mi100_hpe.yaml
```diff
@@ -37,6 +37,7 @@ superbench:
...
     kernel-launch:
       <<: *default_local_mode
+      timeout: 120
```
cp5555, line 40: Besides, why only kernel-launch only? If we cannot make sure the setting, we can revise it in the next PR~~

cp5555 (approved):

superbench/config/amd_mi100_hpe.yaml
guoshzhao: Just the PR description, I just take kernel-launch as example to show how to add timeout.

superbench/config/amd_mi100_hpe.yaml
guoshzhao: Good idea, will update the docs. I thought amd GPU should have slower kernel launch speed, so enlarge it to 180 seconds.

superbench/config/amd_mi100_hpe.yaml
cp5555: I think we should not change the default on this.

superbench/runner/runner.py
```diff
@@ -102,12 +102,13 @@ def __get_enabled_benchmarks(self):
                 return list(self._sb_config.superbench.enable)
         return [k for k, v in self._sb_benchmarks.items() if v.enable]
 
-    def __get_mode_command(self, benchmark_name, mode):
+    def __get_mode_command(self, benchmark_name, mode, timeout=None):
```
abuccts, line 105: change default value to 0?

superbench/runner/runner.py
```diff
@@ -355,14 +359,20 @@ def _run_proc(self, benchmark_name, mode, vars):
...
                 'docker exec sb-workspace bash -c '
                 "'set -o allexport && source sb.env && set +o allexport && {command}'"
-            ).format(command=self.__get_mode_command(benchmark_name, mode))
+            ).format(command=self.__get_mode_command(benchmark_name, mode, timeout))
         )
         if mode.name == 'mpi':
             ansible_runner_config = self._ansible_client.update_mpi_config(ansible_runner_config)
+
+        if timeout is not None:
+            ansible_runner_config['timeout'] = timeout
```
abuccts, line 374: no need to check?

superbench/runner/runner.py
```diff
@@ -116,6 +117,9 @@ def __get_mode_command(self, benchmark_name, mode):
             name=benchmark_name,
             output_dir=self._sb_output_dir,
         )
+        if timeout is not None:
+            exec_command = 'timeout {timeout} {command}'.format(timeout=timeout, command=exec_command)
```
abuccts, line 121: do we still need the bash timeout if you add it in ansible?

abuccts: does it have some side effects? e.g., cannot pass env correctly

superbench/config/amd_mi100_hpe.yaml
guoshzhao: Your suggestion is that we don't change the config files in this PR?

superbench/runner/runner.py
guoshzhao: If we don't set the timeout in config file, it will be `None`. I think use `None` as default is better?

superbench/runner/runner.py
guoshzhao: I tested that ansible timeout only early stop the server side, 'sb exec' processes are still running on client side. We need to terminate all the processes manually.
Didn't find the pass env issue, tested some E2E model benchmarks and microbenchmarks, it works well.

superbench/runner/runner.py
guoshzhao: Removed.

dockerfile/cuda11.1.1.dockerfile
```diff
@@ -38,6 +38,7 @@ RUN apt-get update && \
     openssh-client \
     openssh-server \
     pciutils \
+    timelimit \
```
abuccts, line 41: is it different from timeout in coreutils package?

abuccts (approved):

PR #295, opened 2022-01-28
Config - Support customized env for all modes

abuccts:
Support customized env for all modes in configuration.

### DIFF
--- a/superbench/runner/runner.py
+++ b/superbench/runner/runner.py
@@ -102,12 +102,13 @@ class SuperBenchRunner():
                 return list(self._sb_config.superbench.enable)
         return [k for k, v in self._sb_benchmarks.items() if v.enable]
 
-    def __get_mode_command(self, benchmark_name, mode):
+    def __get_mode_command(self, benchmark_name, mode, timeout=None):
         """Get runner command for given mode.
 
         Args:
             benchmark_name (str): Benchmark name.
             mode (DictConfig): Runner mode.
+            timeout (int): The timeout value in seconds.
 
         Return:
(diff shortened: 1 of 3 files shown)


cp5555 (approved):

superbench/runner/runner.py
```diff
@@ -359,11 +359,15 @@ def _run_proc(self, benchmark_name, mode, vars):
         logger.info('Runner is going to run %s in %s mode, proc rank %d.', benchmark_name, mode.name, mode.proc_rank)
 
         timeout = self._sb_benchmarks[benchmark_name].timeout
+        env_list = '--env-file sb.env'
+        for k, v in mode.env.items():
+            if v is None:
+                continue
+            env_list += ' -e {k}={v}'.format(k=k, v=str(v).format(proc_rank=mode.proc_rank, proc_num=mode.proc_num))
```
guoshzhao, line 366: why here always need to format `proc_rank=mode.proc_rank, proc_num=mode.proc_num`?

guoshzhao (approved):

yzygitzh (approved):

yukirora:
not related to this pr, but it seems there is a bug on line153 in runner.py, pls change if v to if v is None, since v maybe 0

abuccts:
@yukirora environment variable is supposed to be string, not int

superbench/runner/runner.py
abuccts: to support variables like
```yaml
env:
  CUDA_VISIBLE_DEVICES: {proc_rank}
```

PR #299, opened 2022-01-30
Benchmarks: Add Feature - Add GDR-only nccl-tests for Nvidia machines

yzygitzh:
**Description**
This commit adds GDR-only nccl-tests for Nvidia machines. Also bump NCCL to v2.10.3-1 to achieve peak performance in this test.

### DIFF
--- a/dockerfile/cuda11.1.1.dockerfile
+++ b/dockerfile/cuda11.1.1.dockerfile
@@ -94,7 +94,7 @@ RUN cd /tmp && \
 RUN cd /tmp && \
     git clone -b v2.10.3-1 https://github.com/NVIDIA/nccl.git && \
     cd nccl && \
-    make -j src.build && \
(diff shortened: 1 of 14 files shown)


abuccts (approved):

cp5555 (approved):

dockerfile/cuda11.1.1.dockerfile
```diff
@@ -7,7 +7,7 @@ FROM nvcr.io/nvidia/pytorch:20.12-py3
 # NVIDIA:
 #   - CUDA: 11.1.1
 #   - cuDNN: 8.0.5
-#   - NCCL: bootstrap_tag
+#   - NCCL: v2.10.3-1
```
cp5555, line 10: If we change the default NCCL version, all baseline need to be re-tested. @abuccts

PR #305, opened 2022-02-09
Benchmarks: Revise Code - Add support for pytorch>=1.9.0 of init_process_group

yukirora:
**Description**
Add support for pytorch>=1.9.0 of init_process_group.

**Major Revision**
- Use PrefixStore(TCPStore) to init_process_group manully for each model run


superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -241,3 +259,20 @@ def _cal_params_count(self):
...
+
+        Args:
+            x (str): pytorch version
+            y (str): pytorch version
+
+        Returns:
+            bool: True if x >= y, otherwise False
+        """
+        x = [int(i) for i in re.match(r'\d+.\d+.\d+', x).group().split('.')]
+        y = [int(i) for i in re.match(r'\d+.\d+.\d+', y).group().split('.')]
```
abuccts, line 274: I don't think this function can compare `1.9.0a0` and `1.9.0` correctly

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -66,7 +69,22 @@ def _init_distributed_setting(self):
                     )
                     return False
 
-                torch.distributed.init_process_group(backend=self._args.distributed_backend.value)
+                if self.version_larger(str(torch.__version__), '1.9.0'):
```
abuccts, line 72:
```py
from packaging import version
if version.parse(torch.__version__) >= version.parse('1.9.0'):
   ...
```

abuccts: seems you need to compare with '1.9.0a0' instead of '1.9.0'?

or could you just test the api changes instead of comparing version? for example (not verified)
```py
try:
    from torch.distributed import elastic
    # >= 1.9.0a0
except ImportError:
    # < 1.9.0a0
```

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -66,7 +69,22 @@ def _init_distributed_setting(self):
...
+                    port = int(os.environ['MASTER_PORT']) + 1
+                    addr = os.environ['MASTER_ADDR']
+                    rank = int(os.environ['RANK'])
+                    world_size = int(os.environ['WORLD_SIZE'])
+                    logger.debug('ip:{},port:{},rank:{},world:{}'.format(addr, str(port), str(rank), str(world_size)))
```
abuccts, line 77: remove "str" in format

superbench/benchmarks/model_benchmarks/pytorch_base.py
yukirora: really thanks, it works, updated to use api to check

superbench/benchmarks/model_benchmarks/pytorch_base.py
yukirora: updated

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -65,8 +67,25 @@ def _init_distributed_setting(self):
                         ' distributed implementation: {}.'.format(self._name, self._args.distributed_impl)
                     )
                     return False
-
-                torch.distributed.init_process_group(backend=self._args.distributed_backend.value)
+                try:
+                    from torch.distributed import elastic    # noqa F401
```
cp5555, line 71: Shall we don't add `#noqa F401`

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -65,8 +67,25 @@ def _init_distributed_setting(self):
                         ' distributed implementation: {}.'.format(self._name, self._args.distributed_impl)
                     )
                     return False
-
-                torch.distributed.init_process_group(backend=self._args.distributed_backend.value)
+                try:
```
cp5555, line 70: It looks like that we prefer this solution. Why not do `try: torch.distributed.init_process_group(backend=self._args.distributed_backend.value)`

superbench/benchmarks/model_benchmarks/pytorch_base.py
yukirora: if not add it, it can not pass lint for 'F401: import but not used'

superbench/benchmarks/model_benchmarks/pytorch_base.py
cp5555: Yes, that's why we should `try` it first.

superbench/benchmarks/model_benchmarks/pytorch_base.py
cp5555: I think you should import it in the beginning, not in this code.

cp5555 (approved):

PR #308, opened 2022-02-10
[hack] running `sb exec` commands locally (no docker)

olehb:
This PR removes docker from `sb exec` commands. This is needed when running `sb run` from inside a docker container.

### DIFF
--- a/superbench/cli/_handler.py
+++ b/superbench/cli/_handler.py
@@ -294,5 +294,5 @@ def run_command_handler(
         config_override=config_override,
     )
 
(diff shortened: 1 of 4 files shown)


abuccts:
Hi @olehb, we have supported the "running `sb run` inside a Docker container" feature in #358. You could leverage `--no-docker` argument in `sb run` to use it, here's an example to quick try:
```sh
$ docker run -it --privileged --net=host --ipc=host --gpus=all superbench/superbench:latest bash
$ sb run --no-docker -l localhost -C superbench.enable=kernel-launch
```

I will close this draft PR, pls let us know if you have other requirements or issues.

PR #311, opened 2022-02-15
Config - Add T4 configurations with inference benchmark

abuccts:
Add T4 configurations with inference benchmark.


cp5555 (approved):

yukirora (approved):

superbench/config/azure/inference/nc64as_t4_v3.yaml
```diff
@@ -0,0 +1,317 @@
...
+      model_action:
+        - inference
+      pin_memory: yes
+  benchmarks:
+    kernel-launch:
+      <<: *default_local_mode
+    gemm-flops:
+      <<: *default_local_mode
+      parameters:
+        precision: [fp32, fp16, fp16_tc]
```
yukirora, line 54: why we don't test int8

abuccts: no TF32 support in Turing, INT8 is not added before.
https://github.com/microsoft/superbenchmark/blob/97ed12f97ff9f201ea375bfe778e2f96170d66e8/superbench/benchmarks/micro_benchmarks/cuda_gemm_flops_performance.py#L29-L34

PR #320, opened 2022-03-01
Analyzer: Add feature - Add result summary in excel,md,html format

yukirora:
**Description**
Add result summary in excel,md,html format.

**Major Revision**
- Add ResultSummary class to support result summary in excel,md,html format.
- Abstract RuleBase class for common-used functions in DataDiagnosis and ResultSummary.


cp5555 (approved):

superbench/analyzer/file_handler.py
```diff
@@ -221,3 +222,29 @@ def output_lines_in_html(lines, output_path):
...
+
+    Args:
+        ws (worksheet): the worksheet of the excel to process
+        row (int): the max row index to merge
+        column (int): the index of the column to merge
+    """
+    dict_from = {}
+    aligncenter = Alignment(horizontal='center', vertical='center')
+    # record continuous row index (start, end) with the same content
+    for row_index in range(1, row + 1):
```
guoshzhao, line 238: Does worksheet always have the title row?

superbench/analyzer/file_handler.py
```diff
@@ -221,3 +222,29 @@ def output_lines_in_html(lines, output_path):
             f.writelines(html_str)
     except Exception as e:
         logger.error('FileHandler: html_data_output - {}'.format(str(e)))
+
+
+def merge_column_in_excel(ws, row, column):
```
guoshzhao, line 227: So the input data should make sure that the same contents should be continuous? Should we add some check?

superbench/analyzer/result_summary.py
```diff
@@ -0,0 +1,244 @@
...
+            logger.log_and_raise(exception=Exception, msg='{} lack of metrics'.format(name))
+        if 'statistics' not in rule:
+            logger.log_and_raise(exception=Exception, msg='{} lack of function'.format(name))
+        # convert single statistic str to list
+        if not isinstance(rule['statistics'], list):
+            rule['statistics'] = [rule['statistics']]
+        # check statistics format, should be SummaryType or p\d\d?
+        for function in rule['statistics']:
+            if not (re.fullmatch(r'p\d\d?', function) or isinstance(SummaryType(function), SummaryType)):
+                logger.log_and_raise(exception=Exception, msg='{} invalid function name'.format(name))
```
guoshzhao, line 42: Here should be `format(function)` not `format(name)`?
But if `SummaryType(function)` raise Exception, this line will not get the chance to execute.

superbench/analyzer/result_summary.py
```diff
@@ -0,0 +1,244 @@
...
+                summary_df_of_rule = data_analysis.round_significant_decimal_places(
+                    summary_df_of_rule, round, list(summary_df_of_rule.columns)
+                )
+            # format summary_df of a rule to list of lines
+            summary_lines_of_rule = self._format_summary_of_rule(category, summary_df_of_rule)
+            summary[category] = summary_lines_of_rule
+
+        return summary
+
+    def gen_md_lines(self, summary):
```
guoshzhao, line 158: why some function name use `generate`, and other use `gen`?

superbench/analyzer/file_handler.py
yukirora: no, it doesn't have title, 1 is the first index in worksheet

superbench/analyzer/file_handler.py
yukirora: I think this function is designed to merge the cells for continuous same contents rather than merge cells for same contents. if the contents are not continuous but the same, it means they don't want to be merged, that depends on how the input data want to display.

superbench/analyzer/result_summary.py
yukirora: unify to generate

superbench/analyzer/result_summary.py
yukirora: 1. the name is a tip for rule name only, but I can add function name in the end. 
2. yes, add a try for that, and

guoshzhao (approved):

hpourreza:
Thanks @yukirora. Could you please write a readme here, https://github.com/microsoft/superbenchmark/tree/main/superbench/analyzer, for example on what the expected input and the produced output will look like? Maybe you already have this readme.

PR #330, opened 2022-03-16
Bug: Benchmarks - remove fp16  samples type converting time for cnn and lstm models

yukirora:
**Description**
Remove fp16  samples type converting time for cnn and lstm models.

### DIFF
--- a/superbench/benchmarks/model_benchmarks/pytorch_cnn.py
+++ b/superbench/benchmarks/model_benchmarks/pytorch_cnn.py
@@ -134,8 +134,8 @@ class PytorchCNN(PytorchBase):
             self._model.eval()
             while True:
                 for idx, sample in enumerate(self._dataloader):
-                    start = time.time()
                     sample = sample.to(dtype=getattr(torch, precision.value))
+                    start = time.time()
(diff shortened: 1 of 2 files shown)


superbench/benchmarks/model_benchmarks/pytorch_cnn.py
```diff
@@ -134,8 +134,8 @@ def _inference_step(self, precision):
             self._model.eval()
             while True:
                 for idx, sample in enumerate(self._dataloader):
-                    start = time.time()
                     sample = sample.to(dtype=getattr(torch, precision.value))
```
abuccts, line 137: can we just remove this line, and use the corresponding precision (i.e., `self._args.precision`) instead of fp32 when generating dataset? so it's consistent with other transformer models.
https://github.com/microsoft/superbenchmark/blob/ff51a3cee9aa986616bfab4071e9f9ccec81b31b/superbench/benchmarks/model_benchmarks/pytorch_cnn.py#L57

guoshzhao (approved):

cp5555 (approved):

PR #333, opened 2022-03-28
Benchmarks: Add Feature - Provide option to save raw data into file.

guoshzhao:
**Description**
Use config `log_raw_data` to control whether log the raw data into file or not. The default value is `no`. We can set it as `yes` for some particular benchmarks to save the raw data into file, such as NCCL/RCCL test.

### DIFF
--- a/docs/superbench-config.mdx
+++ b/docs/superbench-config.mdx
@@ -336,8 +336,9 @@ A list of models to run, only supported in model-benchmark.
 
 Parameters for benchmark to use, varying for different benchmarks.
 
-There have two common parameters that work for all benchmarks:
+There have three common parameters for all benchmarks:
 * run_count: how many times do user want to run this benchmark, default value is 1.
+* duration: the elapsed time of benchmark in seconds. It can work for all model-benchmark. But for micro-benchmark, benchmark authors should consume it by themselves.
 * log_raw_data: log raw data into file instead of saving it into result object, default value is `False`.  Benchmarks who have large raw output may want to set it as `True`, such as `nccl-bw`/`rccl-bw`.
 
 For Model-Benchmark, there have some parameters that can control the elapsed time.
--- a/docs/superbench-config.mdx
+++ b/docs/superbench-config.mdx
@@ -209,6 +209,7 @@ ${benchmark_name}:
   parameters:
     run_count: int
     duration: int
+    log_raw_data: bool
     ${argument}: bool | str | int | float | list
 ```
 
@@ -224,6 +225,7 @@ model-benchmarks:${annotation}:
   parameters:
     run_count: int
     duration: int
+    log_raw_data: bool
     num_warmup: int
     num_steps: int
     sample_count: int
@@ -333,6 +335,17 @@ A list of models to run, only supported in model-benchmark.
 ### `parameters`
 
 Parameters for benchmark to use, varying for different benchmarks.
+
+There have two common parameters that work for all benchmarks:
+* run_count: how many times do user want to run this benchmark, default value is 1.
+* log_raw_data: log raw data into file instead of saving it into result object, default value is `False`.  Benchmarks who have large raw output may want to set it as `True`, such as `nccl-bw`/`rccl-bw`.
+
+For Model-Benchmark, there have some parameters that can control the elapsed time.
+* duration: the elapsed time of benchmark in seconds.
+* num_warmup: the number of warmup step.
+* num_steps: the number of test step.
+
+If `duration > 0` and `num_warmup + num_steps > 0`, then benchmark will take the least as the elapsed time. Otherwise only one of them will take effect.
 
 ## `Mode` Schema
 
--- a/superbench/benchmarks/model_benchmarks/model_base.py
+++ b/superbench/benchmarks/model_benchmarks/model_base.py
@@ -400,8 +400,8 @@ class ModelBenchmark(Benchmark):
         # The unit of step time is millisecond, use it to calculate the throughput with the unit samples/sec.
         millisecond_per_second = 1000
         throughput = [millisecond_per_second / step_time * self._args.batch_size for step_time in step_times]
-        self._result.add_raw_data(metric_s, step_times)
-        self._result.add_raw_data(metric_t, throughput)
+        self._result.add_raw_data(metric_s, step_times, self._args.log_raw_data)
+        self._result.add_raw_data(metric_t, throughput, self._args.log_raw_data)
 
         if model_action == ModelAction.TRAIN:
             if not self._sync_result(step_times):
--- a/superbench/benchmarks/result.py
+++ b/superbench/benchmarks/result.py
@@ -3,6 +3,7 @@
 
 """A module for unified result of benchmarks."""
 
+import os
 import json
 from enum import Enum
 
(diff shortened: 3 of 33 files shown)


superbench/benchmarks/result.py
```diff
@@ -64,9 +65,22 @@ def add_raw_data(self, metric, value):
...
+            logger.error(
+                'EXPORT_RAW_DATA is invalid: {}'.format(os.getenv('EXPORT_RAW_DATA', '1'))
+            )
+
+        if export_raw_data != 0:
+            if metric not in self.__raw_data:
+                self.__raw_data[metric] = list()
+            self.__raw_data[metric].append(value)
+        else:
+            with open(os.path.join(os.getcwd(), 'rawdata.log'), 'a') as f:
```
abuccts, line 81: the logic is opposite here? when `export_raw_data != 0` export/save it to log file?

superbench/benchmarks/result.py
```diff
@@ -64,9 +65,22 @@ def add_raw_data(self, metric, value):
             )
             return False
 
-        if metric not in self.__raw_data:
-            self.__raw_data[metric] = list()
-        self.__raw_data[metric].append(value)
+        export_raw_data = 1
+        try:
+            export_raw_data = int(os.getenv('EXPORT_RAW_DATA', '1'))
```
abuccts, line 70: maybe add a prefix "SB_BENCHMARK_" for the env?

superbench/benchmarks/result.py
guoshzhao: `export` means exporting to the caller. If save to log, we can use save_raw_data.

superbench/benchmarks/result.py
guoshzhao: OK.

cp5555: Agree!

tests/benchmarks/test_result.py
```diff
@@ -20,12 +22,24 @@ def test_add_raw_data():
     result = BenchmarkResult('model', BenchmarkType.MODEL, ReturnCode.SUCCESS)
     result.add_raw_data('metric1', [1, 2, 3])
     result.add_raw_data('metric1', [4, 5, 6])
-
     assert (result.raw_data['metric1'][0] == [1, 2, 3])
     assert (result.raw_data['metric1'][1] == [4, 5, 6])
     assert (result.type == BenchmarkType.MODEL)
     assert (result.return_code == ReturnCode.SUCCESS)
 
+    assert (int(os.getenv('EXPORT_RAW_DATA', '1')) == 1)
```
cp5555, line 30: Why we test the default value?

superbench/benchmarks/result.py
cp5555: If so, we export_raw_data is related to rawdata.log?

superbench/benchmarks/result.py
```diff
@@ -64,9 +65,22 @@ def add_raw_data(self, metric, value):
             )
             return False
 
-        if metric not in self.__raw_data:
-            self.__raw_data[metric] = list()
-        self.__raw_data[metric].append(value)
+        export_raw_data = 1
```
cp5555, line 68: Why do we need to init this? If the environment doesn't work, we will error directly

guoshzhao: I think this environment needn't to break the whole execution, so just log the error message and use the default value.

superbench/benchmarks/result.py
guoshzhao: Didn't get your point, there have two thing to determine,
1. the variable name, export_raw_data or save_raw_data?

which one do you prefer?
```
if export_raw_data:
    export to caller
else:
    save to file
```

or

```
if save_raw_data:
    save to file
else:
    export to caller
```

2. the default value, is 'export raw data to caller' or 'save the raw data to file'?

Currently, I used export_raw_data, and the default value is set as 1 (true) to export the raw data to caller.

tests/benchmarks/test_result.py
guoshzhao: If the default value is not as expected, other test cases (check the benchmark results) may also fail. And one check here to directly expose it.

cp5555: Therefore, we should save the environment setting first and set it to our expected value. At the end, we will resume the environment setting.

guoshzhao: I mean that I want to make sure the default value is '1', so I add one directly check here.

superbench/benchmarks/result.py
guoshzhao: Use parameter to replace env variable.

tests/benchmarks/test_result.py
guoshzhao: removed.

superbench/benchmarks/result.py
guoshzhao: removed.

cp5555 (approved):

superbench/benchmarks/result.py
cp5555: It has been removed

superbench/benchmarks/result.py
cp5555: Solved

superbench/benchmarks/model_benchmarks/model_base.py
```diff
@@ -400,8 +400,8 @@ def __process_model_result(self, model_action, precision, step_times):
         # The unit of step time is millisecond, use it to calculate the throughput with the unit samples/sec.
```
cp5555, line 400: Would you please add new parameters in document as well? I think it should be in API.

guoshzhao: Thanks, have updated the docs.

docs/superbench-config.mdx
```diff
@@ -209,6 +209,7 @@ ${benchmark_name}:
   parameters:
     run_count: int
     duration: int
```
abuccts, line 211: do we need to remove `duration` in micro-benchmark?

abuccts (approved):

superbench/executor/executor.py
```diff
@@ -200,6 +200,7 @@ def exec(self):
             benchmark_config = self._sb_benchmarks[benchmark_name]
             benchmark_results = list()
             self.__create_benchmark_dir(benchmark_name)
+            os.chdir(self.__get_benchmark_dir(benchmark_name))
```
abuccts, line 203: may need to change back in the end of loop to pass tests..

docs/superbench-config.mdx
guoshzhao: I think we can keep it in case some micro-benchmark can use it to control the execution time by itself, what do you think?

cp5555 (approved):

docs/superbench-config.mdx
```diff
@@ -334,6 +336,17 @@ A list of models to run, only supported in model-benchmark.
 
 Parameters for benchmark to use, varying for different benchmarks.
 
+There have two common parameters that work for all benchmarks:
+* run_count: how many times do user want to run this benchmark, default value is 1.
+* log_raw_data: log raw data into file instead of saving it into result object, default value is `False`.  Benchmarks who have large raw output may want to set it as `True`, such as `nccl-bw`/`rccl-bw`.
+
+For Model-Benchmark, there have some parameters that can control the elapsed time.
+* duration: the elapsed time of benchmark in seconds.
```
abuccts, line 344: so shall we move `duration` from "for mode-benchmark" to "for all benchmarks"?

abuccts: do we support it in code for micro-benchmark currently?

superbench/executor/executor.py
guoshzhao: Sure

abuccts (approved):

PR #342, opened 2022-04-14
Bug -  Fix bugs in sync results on root rank for e2e model benchmarks

yukirora:
**Description**
Fix bugs in sync results on root rank for e2e model benchmarks.

Bugs:
 - results were not changed to sync results (grammer)
 - sync results not applyed to all ranks but only root rank
 - output result on local_rank 0 not global root rank


superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -206,8 +207,10 @@ def _sync_result(self, result):
                     tensor = torch.as_tensor(result).cuda()
                 else:
                     tensor = torch.as_tensor(result)
-                torch.distributed.reduce(tensor, 0, op=torch.distributed.ReduceOp.MAX)
-                result = tensor.tolist()
+                torch.distributed.all_reduce(tensor, op=torch.distributed.ReduceOp.MAX)
+                sync_result = tensor.tolist()
+                for i in range(len(result)):
+                    result[i] = sync_result[i]
```
abuccts, line 213: why not use `result = tensor.tolist()` directly?

yukirora: This is a grammer issue about using Pointer in function params.
Guoshuai treats the function param 'result' as Pointer, and he expected the content of 'result' should be changed to sync_results after this function return.
However, if using 'result = tensor.tolist()', the address of  'result' changes, and after the function returns, the content of ’result‘ in the function that called this func does not change at all, resuting in this function and his pr not take effect(the final results not sync) in fact.

guoshzhao (approved):

abuccts (approved):

PR #344, opened 2022-04-14
Analyzer - Support regex in annotations of benchmark naming for metrics in rules

yukirora:
**Description**
 Support regex in annotations of benchmark naming for metrics in rules.
For example:      
metrics:   
 \- model-benchmarks:resnet50:float/.\*/fp16_train_throughput' 
 -> 
 \- 'model-benchmarks:.\*/.\*/fp16_train_throughput'


superbench/analyzer/rule_base.py
```diff
@@ -32,6 +32,9 @@ def _get_metrics_by_benchmarks(self, metrics_list):
                 logger.warning('RuleBase: get_metrics_by_benchmarks - {} does not have benchmark_name'.format(metric))
             else:
                 benchmark = metric.split('/')[0]
+                # support annotations in benchmark naming
+                if ':' in benchmark:
+                    benchmark = metric.split(':')[0]
```
abuccts, line 37: how will you differentiate benchmarks with same name, e.g., `model-benchmarks:resnet` and `model-benchmarks:bert`, `nccl-bw:default` and `nccl-bw:gdr-only`?

yukirora: I think I don't need to differentiate benchmarks in this func, that depends on how you write rules.
This function is only use to get the real name of benchmark which will be used as keys to re-organize the metrics in raw data and the metrics specified in rules. 
Then the keys can be used to reduce search space when match the metrics specified in rules with regex with the metric names in raw data

yukirora: My goal of this change to support regex for annotations to write metrics in the rules, previously we can not use regex before '/' in the metric, so for example, when we write a rule for result summary, if we change the annotations, we must always change the metric name in the rule.

cp5555 (approved):

abuccts (approved):

PR #345, opened 2022-04-14
Bug - Support no matching rules and unify the output name in result_summary

yukirora:
**Description**
Support no matching rules and unify the output name in result_summary

**Major Revision**
- Support rule with no matched metrics in result summary
- Unify output file name to 'results-summary'


superbench/analyzer/result_summary.py
```diff
@@ -129,6 +129,10 @@ def _generate_summary(self, round):
         """
         summary = {}
         for rule in self._sb_rules:
+            # skip the rule with no matched metrics
+            if len(self._sb_rules[rule]['metrics']) == 0:
+                logger.warning('ResultSummary: No matched metrics - {}'.format(rule))
+                continue
```
cp5555, line 135: I cannot get the point. Why do we need to tolerate this? Warning is OK.

cp5555: I think it's a bug actually. We should support it.

cp5555 (approved):

abuccts (approved):

PR #355, opened 2022-05-30
Analyzer - Fix bugs in data diagnosis

yukirora:
**Description**
Fix bugs in data diagnosis.

**Major Revision**
- add support to get baseline of the metric which uses custom benchmark naming with ':' like 'nccl-bw:default/allreduce_8_bw:0'
- save raw data of all metrics rather than metrics defined in diagnosis_rules.yaml when output_all is True
- fix bug of using wrong column index when applying format(red color and percentile) in the excel


tests/data/diagnosis_summary.json
```diff
@@ -1,5 +1,298 @@
 [
```
cp5555, line 1: What's the goal to revise it?

cp5555 (approved):

tests/data/diagnosis_summary.json
yukirora: to pass the test for ‘save raw data of all metrics rather than metrics defined in diagnosis_rules.yaml when output_all is True’

abuccts (approved):

PR #365, opened 2022-06-24
CLI - Support SKU auto detect if running on Azure VM

abuccts:
Support SKU auto detect and using corresponding benchmark config if running on Azure VM.


superbench/common/utils/file_handler.py
```diff
@@ -57,16 +57,26 @@ def create_sb_output_dir(output_dir=None):
...
     Returns:
         OmegaConf: Config object, None if file does not exist.
     """
-    default_config_file = Path(__file__).parent / '../../config/default.yaml'
-    p = Path(config_file) if config_file else default_config_file
+    p = Path(str(config_file))
+    if not config_file:
+        config_path = Path(__file__).parent / '../../config'
+        p = config_path / 'default.yaml'
+        vm_size = get_vm_size().replace('Standard_', '').lower()
```
cp5555, line 72: Maybe we don't need to replace?

superbench/common/utils/file_handler.py
```diff
@@ -57,16 +57,26 @@ def create_sb_output_dir(output_dir=None):
...
+        if vm_size:
+            logger.info('Detected Azure SKU standard_%s.', vm_size)
+            for config in (config_path / 'azure').glob('**/*'):
+                if config.name.startswith(vm_size):
+                    p = config
+                    break
+        logger.info('No benchmark config provided, using config file %s.', str(p))
```
cp5555, line 79: Maybe we cannot find the right configuration. Shall we output the error?

abuccts: we can only cover a few common SKUs, so it's just in best effort
if the SKU is detected but there's no config for it, the following info log will be shown instead of error:
```
[INFO] Detected Azure SKU ...
[INFO] ... using config file default.yaml
```

superbench/common/utils/file_handler.py
abuccts: the prefix is duplicate in all skus, and current [config files](https://github.com/microsoft/superbenchmark/tree/main/superbench/config/azure/inference) do not include it

cp5555: Yes, we should replace all current config name.

superbench/common/utils/file_handler.py
cp5555: Really confused here. Do we support default.yaml? It's strange for SuperBench has default.yaml. User should specify the yaml, otherwise, it can do auto-detection. If there is no support, we should print error

yzygitzh (approved):

superbench/common/utils/file_handler.py
abuccts: updated to keep "standard_" prefix

superbench/common/utils/file_handler.py
abuccts: > Do we support default.yaml?

yes

here's how the config is decided:
```
if config is specified:
    use provided config
else
    use default.yaml
    if sku can be detected and it happens to have a corresponding config:
        use sku.yaml
```
this PR only adds the "sku can be detected" feature, it will try to find a sku.yaml but no guarantee, for other cases (not on azure, unknow sku, etc.) default.yaml is stilled used as what previous code did
if we decide to remove default.yaml, it will be a breaking change and makes --config argument required instead of optional

cp5555 (approved):

PR #370, opened 2022-07-01
Fix issues in ib validation benchmark

abuccts:
Fix several issues in ib validation benchmark:
* continue running when timeout in the middle, instead of aborting whole mpi process
* make timeout parameter configurable, set default to 120 seconds
* avoid mixture of stdio and iostream when print to stdout
* set default message size to 8M which will saturate ib in most cases
* fix hostfile path issue so that it can be auto found in different cases


superbench/runner/playbooks/check_env.yaml
```diff
@@ -26,7 +26,7 @@
     sb_nodes: '{{ hostvars.values() | map(attribute="ansible_hostname") | sort }}'
     sb_env: |
       # sb env
-      SB_WORKSPACE='{{ workspace if skip_docker else "/root" }}'
+      SB_WORKSPACE={{ workspace if skip_docker else '/root' }}
```
cp5555, line 29: Why related to IB?

superbench/runner/runner.py
```diff
@@ -86,7 +86,7 @@ def __validate_sb_config(self):    # noqa: C901
                             'btl_tcp_if_exclude': 'lo,docker0',
                             'coll_hcoll_enable': 0,
                         }
-                    for key in ['PATH', 'LD_LIBRARY_PATH', 'SB_MICRO_PATH']:
+                    for key in ['PATH', 'LD_LIBRARY_PATH', 'SB_MICRO_PATH', 'SB_WORKSPACE']:
```
cp5555, line 89: Why related to IB?

tests/benchmarks/micro_benchmarks/test_ib_traffic_performance.py
```diff
@@ -154,19 +154,19 @@ def test_ib_traffic_performance(self, mock_gpu):
...
-            "/ib_write_bw -F -n 2000 -d mlx5_0 -s 33554432 --report_gbits' --input_config " + \
-            os.getcwd() + '/config.txt'
+        expect_command = "ib_validation --cmd_prefix '" + benchmark._args.bin_dir + \
+            "/ib_write_bw -F -n 2000 -d mlx5_0 -s 33554432 --report_gbits' " + \
+            f'--timeout 120 --hostfile hostfile --input_config {os.getcwd()}/config.txt'
         command = benchmark._bin_name + benchmark._commands[0].split(benchmark._bin_name)[1]
         assert (command == expect_command)
 
-        parameters = '--ib_dev mlx5_0 --iters 2000 --pattern one-to-one --hostfile hostfile --gpu_dev 0'
+        parameters = '--ib_dev mlx5_0 --msg_size 0 --iters 2000 --pattern one-to-one --hostfile hostfile --gpu_dev 0'
```
cp5555, line 163: Do we need to test --msg_size "wrong_string"

superbench/benchmarks/micro_benchmarks/ib_validation_performance/ib_validation_performance.cc
```diff
@@ -301,7 +303,7 @@ float run_cmd(string cmd_prefix, int port, bool server, string hostname) {
 }
 
 // The ranks in vector of (server, client) run commands parallel
-vector<float> run_cmd_parallel(string cmd_prefix, const vector<std::pair<int, int>> &run_pairs_in_parallel,
+vector<float> run_cmd_parallel(string cmd_prefix, int timeout, const vector<std::pair<int, int>> &run_pairs_in_parallel,
```
cp5555, line 306: Why not add the end of this function? What's your thinking here?

superbench/runner/playbooks/check_env.yaml
abuccts: because ib scan benchmark will read host file in this path to generate the pairs, other benchmarks do not need that file

superbench/runner/runner.py
abuccts: same as above

superbench/benchmarks/micro_benchmarks/ib_validation_performance/ib_validation_performance.cc
abuccts: the first two arguments are parsed from boost program options directly. the last one is generated in other functions

tests/benchmarks/micro_benchmarks/test_ib_traffic_performance.py
abuccts: sure, added a false case

jeseszhang1010 (approved):

yukirora (approved):

PR #374, opened 2022-07-13
Add dependencies for topo xml

abuccts:
Add dependencies

* include ndv4-topo.xml in cuda docker images
* require requests version to avoid RequestsDependencyWarning

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -155,6 +155,7 @@ setup(
         'omegaconf==2.0.6',
         'pandas>=1.1.5',
         'pyyaml>=5.3',
+        'requests>=2.28.1',
         'seaborn>=0.11.2',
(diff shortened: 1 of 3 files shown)


cp5555 (approved):

setup.py
```diff
@@ -155,6 +155,7 @@ def run(self):
         'omegaconf==2.0.6',
         'pandas>=1.1.5',
         'pyyaml>=5.3',
+        'requests>=2.28.1',
```
yzygitzh, line 158: Why upgrading this

yzygitzh: Never mind, saw it in description

yzygitzh (approved):

PR #382, opened 2022-08-02
Analyzer: Rename fields in json of data diagnosis to be more readable

yukirora:
**Description**
Rename field in data diagnosis to be more readable.

**Major Revision**
- rename fields according to diagnosis/metric format

**Minor Revision**
- change type of diagnosis/issue_num to be int

### DIFF
--- a/superbench/analyzer/data_diagnosis.py
+++ b/superbench/analyzer/data_diagnosis.py
@@ -232,20 +232,20 @@ class DataDiagnosis(RuleBase):
             data_not_accept_df (DataFrame): defective nodes's detailed information
 
         Returns:
-            DataFrame: all nodes' detailed information inluding ['Accept','#Issues','Category','Issue_Details']
+            DataFrame: all nodes' detailed information inluding ['Accept','Issue_num','Category','Issue_Details']
         """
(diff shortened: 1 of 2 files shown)


superbench/analyzer/data_diagnosis.py
```diff
@@ -232,20 +232,20 @@ def output_all_nodes_results(self, raw_data_df, data_not_accept_df):
             data_not_accept_df (DataFrame): defective nodes's detailed information
 
         Returns:
-            DataFrame: all nodes' detailed information inluding ['Accept','#Issues','Category','Issue_Details']
+            DataFrame: all nodes' detailed information inluding ['Accept','Issue_num','Category','Issue_Details']
```
abuccts, line 235: do you need to make 'Issue_Details' the same format as 'Issue_num'?

abuccts (approved):

superbench/analyzer/data_diagnosis.py
cp5555: Number_Of_Issues

cp5555 (approved):

superbench/analyzer/data_diagnosis.py
yukirora: updated

PR #383, opened 2022-08-02
Gracefully exit when timeout

abuccts:
* Gracefully exit when timeout, add corresponding log and return code.
* Set minimum timeout to 1 minute and enlarge Ansible timeout.

### DIFF
--- a/superbench/runner/runner.py
+++ b/superbench/runner/runner.py
@@ -388,6 +388,9 @@ class SuperBenchRunner():
         logger.info('Runner is going to run %s in %s mode, proc rank %d.', benchmark_name, mode.name, mode.proc_rank)
 
         timeout = self._sb_benchmarks[benchmark_name].timeout
+        if isinstance(timeout, int):
+            timeout = min(timeout, 60)
+
         env_list = '--env-file /tmp/sb.env'
         if self._docker_config.skip:
             env_list = 'set -o allexport && source /tmp/sb.env && set +o allexport'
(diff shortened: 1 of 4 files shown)


superbench/runner/runner.py
```diff
@@ -405,7 +408,9 @@ def _run_proc(self, benchmark_name, mode, vars):
         if mode.name == 'mpi' and mode.node_num != 1:
             ansible_runner_config = self._ansible_client.update_mpi_config(ansible_runner_config)
 
-        ansible_runner_config['timeout'] = timeout
+        if isinstance(timeout, int):
+            # we do not expect timeout in ansible unless subprocess hangs
+            ansible_runner_config['timeout'] = timeout + 300
```
cp5555, line 413: 300 second?

cp5555 (approved):

superbench/runner/runner.py
guoshzhao, line 413: Here the timeout is for async call?

guoshzhao (approved):

superbench/runner/runner.py
abuccts: yes

abuccts: nope, previously "timeout x command" and ansible will both kill the processes at the nearly same time, ansible timeout will not collect error/timeout logs for remote processes

now "timeout x command" will kill the processes and log related info, ansible timeout won't be triggered by default unless bash timeout hang for 300 seconds

PR #396, opened 2022-08-24
Enable latency test in ib traffic validation distributed benchmark

RyoYang:
Enable ib latency test in ib traffic validation distributed benchmark

As Perftest supports CUDA only in BW tests (Refer [perftest source code](https://github.com/linux-rdma/perftest/blob/23f7f8a56892bd9e00b6a4e8bd4bbfdbe122af47/src/perftest_parameters.c#L1652))
Remove `--use_cuda` option from cmd prefix if the test command is bw related

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
+++ b/superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
@@ -296,12 +296,11 @@ class IBBenchmark(MicroBenchmarkWithInvoke):
         msg_size = f'-s {self._args.msg_size}' if self._args.msg_size > 0 else '-a'
         # Add GPUDirect for ib command
         gpu_dev = ''
-        if self._args.gpu_dev is not None:
+        # Perftest supports CUDA/ROCM only in BW tests
+        if self._args.gpu_dev is not None and 'bw' in self._args.command:
             gpu = GPU()
             if gpu.vendor == 'nvidia':
-                # Perftest supports CUDA only in BW tests
-                if 'bw' in self._args.command:
-                    gpu_dev = f'--use_cuda={self._args.gpu_dev}'
+                gpu_dev = f'--use_cuda={self._args.gpu_dev}'
             elif gpu.vendor == 'amd':
                 gpu_dev = f'--use_rocm={self._args.gpu_dev}'
             else:
--- a/superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
+++ b/superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
@@ -296,17 +296,19 @@ class IBBenchmark(MicroBenchmarkWithInvoke):
         msg_size = f'-s {self._args.msg_size}' if self._args.msg_size > 0 else '-a'
         # Add GPUDirect for ib command
         gpu_dev = ''
-        # Perftest supports CUDA/ROCM only in BW tests
-        if self._args.gpu_dev is not None and 'bw' in self._args.command:
-            gpu = GPU()
-            if gpu.vendor == 'nvidia':
-                gpu_dev = f'--use_cuda={self._args.gpu_dev}'
-            elif gpu.vendor == 'amd':
-                gpu_dev = f'--use_rocm={self._args.gpu_dev}'
-            else:
-                self._result.set_return_code(ReturnCode.INVALID_ARGUMENT)
-                logger.error('No GPU found - benchmark: {}'.format(self._name))
-                return False
+        if self._args.gpu_dev is not None:
+            if 'bw' in self._args.command:
+                gpu = GPU()
+                if gpu.vendor == 'nvidia':
+                    gpu_dev = f'--use_cuda={self._args.gpu_dev}'
+                elif gpu.vendor == 'amd':
+                    gpu_dev = f'--use_rocm={self._args.gpu_dev}'
+                else:
+                    self._result.set_return_code(ReturnCode.INVALID_ARGUMENT)
+                    logger.error('No GPU found - benchmark: {}'.format(self._name))
+                    return False
+            elif 'lat' in self._args.command:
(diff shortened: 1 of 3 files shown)


superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
```diff
@@ -299,7 +299,9 @@ def __prepare_general_ib_command_params(self):
         if self._args.gpu_dev is not None:
             gpu = GPU()
             if gpu.vendor == 'nvidia':
-                gpu_dev = f'--use_cuda={self._args.gpu_dev}'
+                # Perftest supports CUDA only in BW tests
+                if 'bw' in self._args.command:
```
abuccts, line 303: maybe move the check to line 299? --use_rocm is not available for latency test neither

superbench/benchmarks/micro_benchmarks/ib_validation_performance/ib_validation_performance.cc
```diff
@@ -260,13 +260,19 @@ void gather_hostnames(vector<string> &hostnames, string filename) {
...
-// TODO: does not work latency tests
 float process_raw_output(string output) {
     float res = -1.0;
     try {
+        string pattern;
         vector<string> lines;
         boost::split(lines, output, boost::is_any_of("\n"), boost::token_compress_on);
-        regex re("\\d+\\s+\\d+\\s+\\d+\\.\\d+\\s+(\\d+\\.\\d+)\\s+\\d+\\.\\d+");
+        if (output.find("BW") != string::npos) {
+            pattern = "\\d+\\s+\\d+\\s+\\d+\\.\\d+\\s+(\\d+\\.\\d+)\\s+\\d+\\.\\d+";
```
abuccts, line 270: could you add two commented lines for example output here? hard to check the regex

abuccts (approved):

superbench/benchmarks/micro_benchmarks/ib_validation_performance/ib_validation_performance.cc
```diff
@@ -260,13 +260,19 @@ void gather_hostnames(vector<string> &hostnames, string filename) {
...
+        } else {
+            pattern = "\\d+\\s+\\d+\\s+\\d+\\.\\d+\\s+\\d+\\.\\d+\\s+\\d+\\.\\d+"
```
yukirora, line 272: this is to parse avg latency, right?

yukirora (approved):

superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
```diff
@@ -299,7 +299,9 @@ def __prepare_general_ib_command_params(self):
         if self._args.gpu_dev is not None:
```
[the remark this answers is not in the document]
RyoYang: If my understanding is correct, this situation can be filtered after removing the condition to this line as mentioned above.

superbench/benchmarks/micro_benchmarks/ib_validation_performance/ib_validation_performance.cc
RyoYang: Yes, as the gap between max and ave is quite significant in latency test, unlike in bw one.
Thus, choosing avg latency instead.

superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
RyoYang: Good point, will update.

superbench/benchmarks/micro_benchmarks/ib_validation_performance/ib_validation_performance.cc
RyoYang: Updated

superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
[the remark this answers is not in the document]
yukirora: I think the warning message is for users to see, not for correctness, so that user know it's a wrong setting.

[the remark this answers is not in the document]
RyoYang: make sense to me, add a  warn log for this, thx.

RyoYang:
@jeseszhang1010 Could you please review the PR? thanks. @abuccts just find no privileges to trigger the pipeline, could you please give me the access?

superbench/benchmarks/micro_benchmarks/ib_validation_performance/ib_validation_performance.cc
```diff
@@ -260,13 +260,28 @@ void gather_hostnames(vector<string> &hostnames, string filename) {
...
+// 595.50 (t_avg)
 float process_raw_output(string output) {
     float res = -1.0;
     try {
+        string pattern;
         vector<string> lines;
         boost::split(lines, output, boost::is_any_of("\n"), boost::token_compress_on);
-        regex re("\\d+\\s+\\d+\\s+\\d+\\.\\d+\\s+(\\d+\\.\\d+)\\s+\\d+\\.\\d+");
+        if (output.find("BW") != string::npos) {
+            pattern = "\\d+\\s+\\d+\\s+\\d+\\.\\d+\\s+(\\d+\\.\\d+)\\s+\\d+\\.\\d+";
```
jeseszhang1010, line 279: this is to parse BW average or BW peak? Looks like it's average, but L270 is peak.

jeseszhang1010 (approved):

superbench/benchmarks/micro_benchmarks/ib_validation_performance/ib_validation_performance.cc
RyoYang: Nice catch, will revise the comment

PR #397, opened 2022-08-25
Enhance parameter parsing to allow spaces in value

abuccts:
Enhance parameter parsing to support string like `"--arg1 value --arg2 'a long string with spaces'"`.


cp5555 (approved):

guoshzhao (approved):

yukirora:
@abuccts  UT failed in your pr,  the issue should be that --json_str '{"a":"b"}' can not be parsed into dict in test of cudnn and clubas functions(line49 in test_cudnn_function.py), because shlex remove all ", it seems the parameter need to be written in  '{\\"a\\":\\"b\\"}'

PR #398, opened 2022-08-25
Update apt packages in dockerfile

abuccts:
Update apt packages in dockerfile, support sudo and ip commands.

### DIFF
--- a/dockerfile/cuda11.1.1.dockerfile
+++ b/dockerfile/cuda11.1.1.dockerfile
@@ -26,6 +26,7 @@ RUN apt-get update && \
     curl \
     dmidecode \
     git \
+    iproute2 \
     jq \
     libaio-dev \
     libcap2 \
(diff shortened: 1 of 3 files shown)


cp5555 (approved):

dockerfile/cuda11.1.1.dockerfile
```diff
@@ -38,6 +39,7 @@ RUN apt-get update && \
     openssh-client \
     openssh-server \
     pciutils \
+    sudo \
```
yukirora, line 42: just curious, why we need sudo inside docker

cp5555: From Liwei's requirement

yzygitzh (approved):

dockerfile/cuda11.1.1.dockerfile
abuccts: `iproute2` is depended by `ibdev2netdev`
as for `sudo`, some environments run containers as a normal user (azure pipeline, some ai platforms) which has privileged permission but cannot perform privileged operations without `sudo`

PR #381, opened 2022-08-01
Auto generate ibstat file for IB performance validation under topo-aware …

RyoYang:
An enhancement for topo-aware IB performance validation #373
This PR will auto-generate a required ibstate file `ib_traffic_topo_aware_ibstat.txt` which is used as input to build a graph

Tested on the 4 A100VM Clsuter.

PR #402, opened 2022-08-30
Auto generate ibstat file by pssh

RyoYang:
**Description**
As MPI can not be inited twice in one same process (by py and c)
Also, MPI env initialized by mpi4py can not be reused in C env

To avoid MPI init issue introduced from mpi4py, rewrite gen_ibstat_file function to generate ibstat file leveraged by pssh

**Major Revision**
- Rewrite gen_ibstat_file function to generate ibstat file leveraged by pssh

**Minor Revision**
- Remove mpi4py dependency

Tested the functionality of topo-aware on 36 nodes cluster


jeseszhang1010 (commented):
Please briefly explain the issue in #381, and why the change in this PR is necessary.

jeseszhang1010:
> [Codecov](https://codecov.io/gh/microsoft/superbenchmark/pull/402?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft) Report

Please double check the codecov report to see if there's real issue or just false positives.

superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
```diff
@@ -254,7 +254,7 @@ def __prepare_config(self):
             if not self._args.hostfile:
                 self._args.hostfile = os.path.join(os.environ.get('SB_WORKSPACE', '.'), 'hostfile')
             with open(self._args.hostfile, 'r') as f:
-                hosts = f.readlines()
+                hosts = list(filter(None, f.read().split('\n')))
```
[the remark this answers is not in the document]
RyoYang: > hosts = f.read().splitlines()?
Addressed

RyoYang:
> Please briefly explain the issue in #381, and why the change in this PR is necessary in the commit message.

Updated in PR description and will be updated to commit message after merging

jeseszhang1010 (approved):
LGTM. I approved it, thank you.
I prefer we test this change on a larger scale, say 32 nodes, and make sure it works as expected.

superbench/common/utils/topo_aware.py
```diff
@@ -35,37 +38,41 @@ def search(self, pattern, string, flags=0):
...
+        # Append each host config into host_config
+        host_config = []
+        for hostx in host_list:
+            hostx_config = config.lookup(hostx)
+            port = hostx_config['port']
+            private_key = hostx_config['identityfile'][0]
+            host_config.append(HostConfig(port=int(port), private_key=private_key))
+        # Fetch ibstat info from each host
+        client = ParallelSSHClient(host_list, host_config=host_config)
+        cmd = r"ibstat | grep -Po 'System image GUID: \K\S+$'"
```
abuccts, line 61: `ibstat` won't work when accelerated networking is enabled when creating VM, pls use ` cat /sys/class/infiniband/*/sys_image_guid | tr -d :` instead

abuccts:
```sh
 parallel-ssh -i -t 5 -p 512 -h hostfile 'cat /sys/class/infiniband/*/sys_image_guid | tr -d :' | grep '[[:alnum:]]\{16\}' | sort | uniq
```
you can run it with parallel-ssh to get all guid

RyoYang: Thanks for the suggestion!
>  parallel-ssh -i -t 5 -p 512 -h hostfile 'cat /sys/class/infiniband/*/sys_image_guid | tr -d :' | grep '[[:alnum:]]\{16\}' | sort | uniq
But if my understanding is correct, this command may cause two issues:
1. All guid with random order which is hard to map the guid with hostname
2. those guid missing the correct prefix can not be used to generate config as the less info to map phy host to guid in topo-aware

RyoYang: udpated, please have a check.

superbench/common/utils/topo_aware.py
```diff
@@ -35,37 +35,23 @@ def search(self, pattern, string, flags=0):
...
-    # Filter the duplicate info
-    ibstat_infos = set(ibstats)
-
-    with ibstate_file_path.open(mode='w') as f:
-        for ibstat_info in ibstat_infos:
-            f.write(ibstat_info)
-    MPI.Finalize()
+    try:
+        pssh_cmd = "pssh -i -t 5 -H '{}' ".format(' '.join(host_list))
+        cmd = "'cat /sys/class/infiniband/*/sys_image_guid | tr -d :' | sed 's/^.*SUCCESS]/VM_hostname/g' | uniq"
```
jeseszhang1010, line 47: Could you elaborate a bit on `^.*SUCCESS]/VM_hostname`

superbench/common/utils/topo_aware.py
jeseszhang1010: Did you find the solution for the above two issues or they are not the issue any more? 
I believe you just removed prefix `0x`when mapping for #2, but not quite sure #1.

RyoYang: Yea, your are right for #2. 
For #1, as the sequential relationship is implied in the output format from pssh is like:
```
[1] 03:03:00 [SUCCESS] <hostname#1>
guid#1
guid#2
guid#3
[2] 03:03:00 [SUCCESS] <hostname#2>
guid#1
guid#2
guid#3
```
so we can just replace the `[1] 03:03:00 [SUCCESS]` to get the correct info

superbench/common/utils/topo_aware.py
RyoYang: sure, will address that

setup.py
```diff
@@ -155,6 +155,7 @@ def run(self):
         'omegaconf==2.0.6',
         'openpyxl>=3.0.7',
         'pandas>=1.1.5',
+        'pssh @ git+https://github.com/lilydjwg/pssh',
```
abuccts, line 158: pls fix the version to https://github.com/lilydjwg/pssh/tree/v2.3.4

superbench/common/utils/topo_aware.py
```diff
@@ -125,8 +111,8 @@ def gen_topo_aware_config(host_list, ibstat_file, ibnetdiscover_file, min_dist,
                     r = quick_regexp()
                     if r.search(r'^(VM_hostname)\s+(.+)', line):
                         vmhost = r.groups[1]
-                    elif r.search(r'^(0x)(.+)', line):
-                        sysimgguid = r.groups[1]
+                    elif r.search(r'^([a-z0-9]{16})$', line):
+                        sysimgguid = r.groups[0]
```
abuccts, line 115: need to exclude all zero lines for accelerated network nic

superbench/common/utils/topo_aware.py
abuccts, line 47: pls remove uniq, it's useless if there's no sort ahead
what if failed vm also prints some lines, it will be mixed with success output?

RyoYang: > pls remove uniq, it's useless if there's no sort ahead
sure, will address
> what if failed vm also prints some lines, it will be mixed with success output?
can we add an error log here if the output contains the fail case?

jeseszhang1010 (commented):
Pls test the change with the VMs with accelerated network

superbench/common/utils/topo_aware.py
abuccts: maybe check "FAILURE" line when parsing the log and ignore following lines until next VM_hostname? you may enhance it in the future instead of this fix

here's a case:
```
VM_hostname node0
000100020003000a
000100020003000b
[2] 03:26:16 [FAILURE] node1 Exited with error code 255
anything in error log
VM_hostname node2
000100020003000d
000100020003000c
```

superbench/common/utils/topo_aware.py
```diff
@@ -125,8 +111,8 @@ def gen_topo_aware_config(host_list, ibstat_file, ibnetdiscover_file, min_dist,
                     r = quick_regexp()
                     if r.search(r'^(VM_hostname)\s+(.+)', line):
                         vmhost = r.groups[1]
-                    elif r.search(r'^(0x)(.+)', line):
-                        sysimgguid = r.groups[1]
+                    elif r.search(r'^([a-z0-9]{16})$', line):
```
abuccts: sys_guid is hex? maybe [a-f0-9]

superbench/common/utils/topo_aware.py
```diff
@@ -35,37 +35,24 @@ def search(self, pattern, string, flags=0):
...
-    ibstat_infos = set(ibstats)
-
-    with ibstate_file_path.open(mode='w') as f:
-        for ibstat_info in ibstat_infos:
-            f.write(ibstat_info)
-    MPI.Finalize()
+    try:
+        pssh_cmd = "pssh -i -t 5 -H '{}' ".format(' '.join(host_list))
+        cmd = "'cat /sys/class/infiniband/*/sys_image_guid" \
+            r"| tr -d :' | sed 's/^.*[[[:upper:]]\{7\}]/VM_hostname/g' | cut -d ' ' -f 1,2"
```
abuccts, line 48: change [[:upper:]]{7} to (SUCCESS|FAILURE)? there're only two options, you may need sed -E

abuccts (approved):

superbench/common/utils/topo_aware.py
RyoYang: addressed

jeseszhang1010 (approved):

superbench/common/utils/topo_aware.py
```diff
@@ -35,37 +36,34 @@ def search(self, pattern, string, flags=0):
...
-    ibstat_infos = set(ibstats)
-
-    with ibstate_file_path.open(mode='w') as f:
-        for ibstat_info in ibstat_infos:
-            f.write(ibstat_info)
-    MPI.Finalize()
+    try:
+        # Only exec on rank0
+        if os.environ.get('OMPI_COMM_WORLD_NODE_RANK') == '0' and os.environ.get('OMPI_COMM_WORLD_LOCAL_RANK') == '0':
+            pssh_cmd = "pssh -i -t 5 -x '-o StrictHostKeyChecking=no' -H '{}' ".format(' '.join(host_list))
```
abuccts, line 49: add "-p 512" for pssh? otherwise it's easy to timeout

superbench/common/utils/topo_aware.py
```diff
@@ -35,37 +36,34 @@ def search(self, pattern, string, flags=0):
...
+            cmd = "'cat /sys/class/infiniband/*/sys_image_guid | tr -d :'" \
+                r"| sed -e 's/^.*\[SUCCESS\]/VM_hostname/g;s/^.*\[FAILURE\]/VM_hostname/g' | cut -d ' ' -f 1,2"
+            output = os.popen(pssh_cmd + cmd).read()
+            # Generate ibstat file
+            ibstate_file_path = Path(ibstat_file)
+            with ibstate_file_path.open(mode='w') as f:
+                f.write(output)
+            scp_cmd = "pscp -t 5 -H '{0}' {1} {1}".format(' '.join(host_list), ibstat_file)
```
abuccts, line 57: -p for pscp

superbench/common/utils/topo_aware.py
RyoYang: addressed

PR #417, opened 2022-10-14
Dockerfile - Update pip install in docker images

yukirora:
**Description**
Update pip install in docker images to update the python package.

### DIFF
--- a/.azure-pipelines/cuda-unit-test.yml
+++ b/.azure-pipelines/cuda-unit-test.yml
@@ -18,8 +18,7 @@ steps:
       echo "##vso[task.prependpath]$HOME/.local/bin"
     displayName: Export path
   - script: |
-      python3 -m pip install --upgrade pip
-      python3 -m pip install --upgrade .[test,nvworker]
+      python3 -m pip install .[test,nvworker]
       make postinstall
     displayName: Install dependencies
   - script: |
--- a/.azure-pipelines/cuda-unit-test.yml
+++ b/.azure-pipelines/cuda-unit-test.yml
@@ -18,7 +18,8 @@ steps:
       echo "##vso[task.prependpath]$HOME/.local/bin"
     displayName: Export path
   - script: |
-      python3 -m pip install -U .[test,nvworker]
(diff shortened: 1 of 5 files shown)


cp5555 (approved):

dockerfile/cuda11.1.1.dockerfile
```diff
@@ -130,6 +130,7 @@ ADD third_party third_party
 RUN make -C third_party cuda
 
 ADD . .
-RUN python3 -m pip install .[nvworker] && \
+RUN python3 -m pip install --upgrade pip && \
+    python3 -m pip install -U .[nvworker] && \
```
abuccts, line 134: why not fix the version for specific package instead? I think pandas has issue because they release a new version recently
upgrading to latest may have more issues for some other packages, like pytorch

cp5555: When I read it, I also have the same feeling on this...

setup.py
```diff
@@ -154,7 +154,7 @@ def run(self):
         'numpy>=1.19.2',
         'omegaconf==2.0.6',
         'openpyxl>=3.0.7',
-        'pandas>=1.1.5',
+        'pandas==1.1.5',
```
abuccts, line 157: seems this line change is also merged in main?

.azure-pipelines/cuda-unit-test.yml
```diff
@@ -11,7 +11,7 @@ pool:
 
 container:
   image: nvcr.io/nvidia/pytorch:20.12-py3
-  options: '-v /var/run/docker.sock:/var/run/docker.sock -v /usr/bin/docker:/usr/bin/docker'
+  options: "-v /var/run/docker.sock:/var/run/docker.sock -v /usr/bin/docker:/usr/bin/docker"
```
abuccts, line 14: is this change necessary?

setup.py
yukirora: oh, if so, we don't need this pr

PR #427, opened 2022-10-28
CLI - Update version to include revision hash and date

abuccts:
Update version to include revision hash and date in `{last tag}+g{git hash}.d{date}` format, here're the examples:
* exact tag: `0.6.0`
* commit after tag: `0.6.0+gcbb1b34`
* commit after tag with local changes: `0.6.0+gcbb1b34.d20221028`

### DIFF
--- a/dockerfile/cuda11.1.1.dockerfile
+++ b/dockerfile/cuda11.1.1.dockerfile
@@ -131,6 +131,6 @@ RUN make -C third_party cuda
 
 ADD . .
 RUN --mount=source=.git,target=.git,type=bind \
-    python3 -m pip install .[nvworker] && \
+    python3 -m pip install --no-cache-dir .[nvworker] && \
     make cppbuild && \
     make postinstall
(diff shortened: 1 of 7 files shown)


cp5555 (approved):

dockerfile/cuda11.1.1.dockerfile
```diff
@@ -130,6 +130,7 @@ ADD third_party third_party
 RUN make -C third_party cuda
 
 ADD . .
-RUN python3 -m pip install .[nvworker] && \
+RUN --mount=source=.git,target=.git,type=bind \
```
cp5555, line 133: I check https://docs.docker.com/storage/volumes/. It looks like
```
RUN --mount 'source=.git,target=.git,type=bind' \
```

abuccts: the link is for `docker run` command args, [this](https://docs.docker.com/engine/reference/builder/#run---mounttypebind) is for dockerfile's `RUN` instruction

it cannot access the build context excluded by dockerignore, so there was an issue

yzygitzh (approved):

PR #434, opened 2022-11-18
Runner - Enable runner expose stdout

RyoYang:
**Description**
Add an option `stdout` to run, enable AnsibleClient can return stdout.

### DIFF
--- a/superbench/runner/ansible.py
+++ b/superbench/runner/ansible.py
@@ -88,6 +88,9 @@ class AnsibleClient():
             logger.warning('Run failed, return code {}.'.format(r.rc))
         if stdout:
             outputs = self._run_stdout_regex.sub('', raw_outputs)
+            # Extra treatment for local condition
+            if ansible_config['host_pattern'] == 'localhost':
+                outputs = outputs.split('rc=0 >>', 1)[1].strip()
             return outputs
         return r.rc
 
--- a/superbench/runner/ansible.py
+++ b/superbench/runner/ansible.py
@@ -5,6 +5,7 @@
 
 import tempfile
 from pathlib import Path
+import re
 
 import ansible_runner
 from ansible.parsing.dataloader import DataLoader
@@ -59,15 +60,17 @@ class AnsibleClient():
                 self._config['cmdline'] += ' --ask-pass --ask-become-pass'
         logger.info(self._config)
 
-    def run(self, ansible_config, sudo=False):    # pragma: no cover
+    def run(self, ansible_config, sudo=False, stdout=False):    # pragma: no cover
         """Run Ansible runner.
 
         Args:
             ansible_config (dict): Ansible config dict.
             sudo (bool): Run as sudo or not. Defaults to False.
+            stdout (bool): return stdout or not. Defaults to False.
 
         Returns:
             int: Ansible return code.
+            outputs: Ansible stdout if stdout is True.
         """
         if sudo:
             logger.info('Run as sudo ...')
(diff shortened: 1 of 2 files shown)


superbench/runner/ansible.py
```diff
@@ -59,27 +60,35 @@ def __init__(self, config):
...
+                raw_outputs = r.stdout.read()
         if r.rc == 0:
             logger.info('Run succeed, return code {}.'.format(r.rc))
         else:
             self.failure_count += 1
             logger.warning('Run failed, return code {}.'.format(r.rc))
+        if stdout:
+            regex = re.compile(r'\x1b(\[.*?[@-~]|\].*?(\x07|\x1b\\))')
+            outputs = regex.sub('', raw_outputs)
+            return outputs
```
yukirora, line 91: change the rc code to others may introduce bug in local mode [runner.py L445](https://github.com/microsoft/superbenchmark/blob/main/superbench/runner/runner.py#L445)

RyoYang: Thanks for the comment. 
Actually, this feature is just for adding a use for others, people can set stdout as True to get output from ansible runner.
If I understand correctly,  I don't think it could introduce an issue for existing code, as stdout's value is always False.

superbench/runner/ansible.py
```diff
@@ -59,27 +60,35 @@ def __init__(self, config):
...
             logger.debug(r.stats)
+            if stdout:
+                raw_outputs = r.stdout.read()
         if r.rc == 0:
             logger.info('Run succeed, return code {}.'.format(r.rc))
         else:
             self.failure_count += 1
             logger.warning('Run failed, return code {}.'.format(r.rc))
+        if stdout:
+            regex = re.compile(r'\x1b(\[.*?[@-~]|\].*?(\x07|\x1b\\))')
```
cp5555, line 89: I think we only need `compile` once.

RyoYang: Do you mean to add the compile into __init__ part to only compile regex only once?

cp5555: Yes, otherwise, it's even slower. Maybe @abuccts has comments

RyoYang: addressed

superbench/runner/ansible.py
```diff
@@ -27,6 +28,7 @@ def __init__(self, config):
             'cmdline': '--forks 128',
         }
         self._head_host = None
+        self._regex = re.compile(r'\x1b(\[.*?[@-~]|\].*?(\x07|\x1b\\))')
```
cp5555, line 31: `_run_stdout_regex`

cp5555 (approved):

superbench/runner/ansible.py
RyoYang: updated

superbench/runner/ansible.py
```diff
@@ -5,6 +5,7 @@
```
cp5555, line 5: Please add test for new feature.

superbench/runner/ansible.py
yukirora: ok, just my concerns, I'm not sure where this feature will be used for, pls at least add test for this feature

superbench/runner/ansible.py
RyoYang: updated

yukirora (approved):

superbench/runner/ansible.py
```diff
@@ -59,27 +61,37 @@ def __init__(self, config):
...
         if r.rc == 0:
             logger.info('Run succeed, return code {}.'.format(r.rc))
         else:
             self.failure_count += 1
             logger.warning('Run failed, return code {}.'.format(r.rc))
+        if stdout:
+            outputs = self._run_stdout_regex.sub('', raw_outputs)
+            # Extra treatment for local condition
+            if ansible_config['host_pattern'] == 'localhost':
+                outputs = outputs.split('rc=0 >>', 1)[1].strip()
```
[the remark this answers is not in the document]
abuccts: seems this output is not designed for parsing, output from different nodes can also mixed and order may not be guaranteed, if you want structured output writing a playbook may be better

superbench/runner/ansible.py
```diff
@@ -59,27 +61,37 @@ def __init__(self, config):
...
+            return outputs
```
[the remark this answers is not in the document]
RyoYang: Good idea! Will draft this PR and use callback instead.

PR #443, opened 2022-12-26
Benchmarks - Enable log flushing for microbenchmark

yukirora:
**Description**

Enbale microbenchmark command run real-time log flushing.

**Major Revision**
- add option for real-time log flushing to stdout in running command of util

### DIFF
--- a/superbench/common/utils/process.py
+++ b/superbench/common/utils/process.py
@@ -4,19 +4,43 @@
 """Process Utility."""
 
 import subprocess
+import sys
+import os
+import shlex
 
 
-def run_command(command):
+def run_command(command, flush_output=False):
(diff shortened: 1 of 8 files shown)


superbench/common/utils/process.py
```diff
@@ -4,19 +4,43 @@
 """Process Utility."""
```
cp5555, line 4: How to enable it by user?

cp5555 (approved):

superbench/common/utils/process.py
yukirora: user can add a parameter to enable it for any benchmark in config, pls check #445

cp5555: Thanks

abuccts (approved):

PR #446, opened 2022-12-26
Benchmarks - Support FP8 in BERT models

abuccts:
Support FP8 in PyTorch BERT models:

* add fp8 hybrid/e4m3/e5m2 in precision arguments
* build BERT encoders with `te.TransformerLayer` to repalce `transformers.BertModel`
* wrap forward steps with fp8 autocast


superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -15,15 +20,37 @@
 
 class BertBenchmarkModel(torch.nn.Module):
     """The BERT model for benchmarking."""
-    def __init__(self, config, num_classes):
+    def __init__(self, config, num_classes, enable_fp8=False):
```
cp5555, line 23: Do we need this? For example, in future, when we add FP4, it will change the interface again. Therefore, shall we try FP8 directly?

cp5555: Can we get the information from configuration.

abuccts: updated

cp5555 (approved):

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -23,7 +28,32 @@ def __init__(self, config, num_classes):
             num_classes (int): The number of objects for classification.
         """
         super().__init__()
-        self._bert = BertModel(config)
+        self._enable_fp8 = False
```
yukirora, line 31: maybe this can be moved into model_base I think? it should be general for all models to support fp8 precision？

yukirora, line 31: maybe this can be moved into model_base I think? it should be general for all models to support fp8 precision？

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -23,7 +28,32 @@ def __init__(self, config, num_classes):
...
+        if hasattr(config, 'precision'):
+            self._enable_fp8 = config.precision.startswith('FP8_')
+            delattr(config, 'precision')
```
yukirora, line 34: maybe this can be moved into model_base I think? it should be general for all models to support fp8 precision？

yukirora (approved):

PR #447, opened 2022-12-27
 Runner - Support `pair-wise` pattern in `mpi` mode

RyoYang:
**Major Revision**
- Extract `pair-wise` pattern from ib-validation benchmark to `gen_traffic_pattern_config.py` for reuse

**Minor Revision**
- Fix typo


superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
```diff
@@ -6,7 +6,7 @@
 import os
```
cp5555, line 6: My suggestion is one PR to revise microbenchmark directly. Without this file, all others look good.

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
RyoYang: Thanks for the suggestion, updated.

abuccts (approved):

PR #452, opened 2022-12-29
Benchmarks: Micro benchmarks - Add correctness check in cublas-function benchmark

yukirora:
**Description**
 Add correctness check in cublas-function benchmark.

**Major Revision**
- add python code of correctness check in cublas-function benchmark and test


abuccts (approved):

superbench/benchmarks/micro_benchmarks/cublas_function.py
```diff
@@ -260,7 +275,7 @@ def _preprocess(self):
             return False
         return True
 
-    def _process_raw_result(self, cmd_idx, raw_output):
+    def _process_raw_result(self, cmd_idx, raw_output):    # noqa: C901
```
cp5555, line 278: Do we have some method to solve C901?

cp5555 (approved):

PR #458, opened 2023-01-03
Runner - Generate host groups file in mpi mode

RyoYang:
**Major Revision**

- Add an option for pattern to generate mpi_pattern.txt file if specified the path.
- In mpi pattern, serial_index and parallel_index will add in each benchmark as environment variables.

**Minor Revision**
- Fix typo


superbench/common/utils/__init__.py
```diff
@@ -11,7 +11,7 @@
 from superbench.common.utils.process import run_command
 from superbench.common.utils.topo_aware import gen_topo_aware_config
 from superbench.common.utils.gen_traffic_pattern_config import (
-    gen_pair_wise_config, gen_traffic_pattern_host_group, gen_ibstat
+    gen_pair_wise_config, gen_traffic_pattern_host_groups, gen_ibstat
```
cp5555, line 14: It's OK to change the function name this time. But it would be great to keep it in future. It's very strange to change the interface time-to-time.

RyoYang: sure, thanks for the suggestion.

superbench/common/utils/gen_traffic_pattern_config.py
```diff
@@ -178,5 +179,14 @@ def gen_traffic_pattern_host_group(host_list, pattern):
...
-    return host_group
+    host_groups = __convert_config_to_host_group(config, host_list)
+    # write traffic pattern host groups to specified path
+    if pattern_config_path:
+        with open(pattern_config_path, 'a') as f:
+            f.write('pattern_type: {}'.format(pattern.type) + '\n')
+            for serial_index, host_group in enumerate(host_groups):
+                for parallel_index, host_list in enumerate(host_group):
+                    exec_index = '{}_{}'.format(serial_index, parallel_index)
+                    f.write("{}: '{}' ".format(exec_index, ' '.join(host_list)))
```
abuccts, line 190: exec_index seems useless, can you keep the same format as current ib validation file? "," to concat different hosts in each group and ";" to concat parallel groups

superbench/runner/runner.py
```diff
@@ -405,6 +405,8 @@ def _run_proc(self, benchmark_name, mode, vars):
             int: Process return code.
         """
         mode.update(vars)
+        if mode.name == 'mpi' and mode.pattern:
+            mode.env.update({'SERIAL_INDEX': mode.serial_index, 'PARALLEL_INDEX': mode.parallel_index})
```
abuccts, line 409: add "SB_MODE_" prefix for env names?

tests/common/test_gen_traffic_pattern_config.py
```diff
@@ -43,7 +45,14 @@ def test_gen_traffic_pattern_host_group(self, tp_hostfile):
             [['node0', 'node5'], ['node6', 'node4'], ['node7', 'node3'], ['node1', 'node2']],
             [['node0', 'node6'], ['node7', 'node5'], ['node1', 'node4'], ['node2', 'node3']]
         ]
-        self.assertEqual(gen_traffic_pattern_host_group(hostx, pattern), expected_host_group)
+        self.assertEqual(gen_traffic_pattern_host_groups(hostx, pattern), expected_host_group)
+        # Test for pattern_config file
+        temp_dir = tempfile.TemporaryDirectory()
```
abuccts, line 50: need to wrap it in `with`, or call `temp_dir.cleanup()` to remove it

tests/common/test_gen_traffic_pattern_config.py
```diff
@@ -43,7 +45,14 @@ def test_gen_traffic_pattern_host_group(self, tp_hostfile):
...
+        test_config_path = '{}/test_pattern_config.txt'.format(temp_dir.name)
```
abuccts, line 51: why not create tempfile.NamedTemporaryFile directly?

superbench/common/utils/gen_traffic_pattern_config.py
yukirora: And single line for each run

tests/common/test_gen_traffic_pattern_config.py
cp5555: If it will be part of results, please add to documentation for this output

superbench/runner/runner.py
```diff
@@ -460,14 +462,19 @@ def run(self):
...
                             host_list = f.read().splitlines()
-                        pattern_hostx = gen_traffic_pattern_host_group(host_list, mode.pattern)
-                        for host_groups in pattern_hostx:
-                            para_rc_list = Parallel(n_jobs=len(host_groups))(
-                                delayed(self._run_proc)
-                                (benchmark_name, mode, vars={
-                                    'proc_rank': 0,
-                                    'host_list': host_group,
-                                }) for host_group in host_groups
+                        host_groups = gen_traffic_pattern_host_groups(host_list, mode.pattern)
```
cp5555, line 465: If you don't need path, why do you support in utils?

superbench/common/utils/gen_traffic_pattern_config.py
RyoYang: updated for both

superbench/runner/runner.py
RyoYang: updated

tests/common/test_gen_traffic_pattern_config.py
RyoYang: updated and add a description in overview.md.

superbench/runner/runner.py
```diff
@@ -460,14 +462,19 @@ def run(self):
...
+                        for serial_index, host_group in enumerate(host_groups):
```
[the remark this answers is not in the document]
RyoYang: I forgot to input the path parameter here so it looks weird. pattern config file will be generated under output folder for each run, and it will be a mandatory operation if user specified pattern in mpi mode, so that they can use that file to diagnose. For the second question, I think the config will not change in each run for all benchmarks as hostfile is not really changed. Duplication config (same pattern for different benchmarks) will be considered in the future.

superbench/runner/runner.py
RyoYang: same with above.

docs/design-docs/overview.md
```diff
@@ -72,6 +72,7 @@ Here're the details about work directory structure for SuperBench Runner.
         │       └── sb-exec.log               # collected SuperBench Executor log
         ├── sb-run.log                        # SuperBench Runner log
         ├── sb.config.yaml                    # SuperBench configuration snapshot
+        ├── pattern_config                    # generated pattern config file for diagnosis
```
cp5555, line 75: Is it optional?

cp5555 (approved):

docs/design-docs/overview.md
yukirora, line 75: generated host file in all mpi patterns of config for diagnosis?

superbench/runner/runner.py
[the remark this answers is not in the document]
RyoYang: Agree with you, as it will be better for helping users to further diagnose. Will update.

docs/design-docs/overview.md
RyoYang: It might be ambiguous if using optional, cause its generation is implied under mpi pattern which is not optional. How about `generated host groups config under specified patterns for diagnosis`

cp5555: I don't think so. For monitor, it's also optional because you can close monitor. it's the same for pattern. Also, if it's for diagnosis, why do we must have it?

docs/design-docs/overview.md
```diff
@@ -72,6 +72,7 @@ Here're the details about work directory structure for SuperBench Runner.
         │       └── sb-exec.log               # collected SuperBench Executor log
         ├── sb-run.log                        # SuperBench Runner log
         ├── sb.config.yaml                    # SuperBench configuration snapshot
+        ├── pattern_config                    # generated host groups config under specified patterns for diagnosis
```
abuccts, line 75: add ".txt", and maybe rename to mpi_pattern?

tests/common/test_gen_traffic_pattern_config.py
```diff
@@ -4,18 +4,24 @@
...
 
 class GenConfigTest(unittest.TestCase):
     """Test the utils for generating config."""
+    @decorator.load_data('tests/data/pattern_config')    # noqa: C901
     @decorator.load_data('tests/data/ib_traffic_topo_aware_hostfile')    # noqa: C901
-    def test_gen_traffic_pattern_host_group(self, tp_hostfile):
+    def test_gen_traffic_pattern_host_group(self, expected_pattern_config, tp_hostfile):
         """Test the function of generating traffic pattern config from specified mode."""
         # Test for all-nodes pattern
+        test_config_file = tempfile.NamedTemporaryFile()
```
abuccts, line 20: add `with` or cleanup it after use

abuccts (approved):

yukirora (approved):

docs/design-docs/overview.md
RyoYang: udpated.

docs/design-docs/overview.md
RyoYang: Add a a bool option `mpi_pattern` under pattern for user, so that they can use this option to get config file for diagnosis if they want,

PR #462, opened 2023-01-13
Bug: Fix bug for incorrect datatype judgement in cublas-function source code

yukirora:
**Description**
Fix bug for incorrect datatype judgement in cublas-function source code.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/cublas_function/cublas_function_helper.h
+++ b/superbench/benchmarks/micro_benchmarks/cublas_function/cublas_function_helper.h
@@ -178,13 +178,13 @@ void from_json(const json &j, CublasFunction &fn) {
         auto datatype = j.at("datatype").get<std::string>();
         fn.set_datatype(datatype);
     } catch (std::exception &e) {
-        throw "invalid datatype in json";
+        fn.set_datatype("unknown");
     }
     try {
         auto use_tensor_core = j.at("use_tensor_core").get<bool>();
         fn.set_use_tensor_core(use_tensor_core);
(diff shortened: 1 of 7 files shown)


rhewett:
This looks good to me.

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/cublas_function/cublas_function_helper.h
```diff
@@ -175,13 +175,25 @@ void from_json(const json &j, CublasFunction &fn) {
...
+        } else {
+            auto datatype = j.at("datatype").get<std::string>();
+            fn.set_datatype(datatype);
+        }
+
+    } catch (std::exception &e) {
+        throw std::runtime_error("datatype is required for cublasGemmEx and cublasGemmStridedBatchedEx");
+    }
+    try {
+        if (str.find("use_tensor_core") == std::string::npos) {
```
rhewett, line 189: I assume this catches the case where the required value is missing?

I think that, if it is required, it should fail if it is missing.

rhewett:
I'm going to approve this as-is.  

I do prefer failing if a required parameter is missing rather than silently going to a default, but I will not block it on that.

rhewett (approved):

superbench/benchmarks/micro_benchmarks/cublas_function/cublas_function_helper.h
yukirora: some functions like pure sgemm do not need this parameter, so this parameter is optional not required in some cases

PR #478, opened 2023-02-13
CI/CD - Upgrade networkx version to fix installation compatibility issue

yukirora:
**Description**
Upgrade networkx version to fix installation compatibility issue.


abuccts (approved):

cp5555 (approved):

.azure-pipelines/cpu-unit-test.yml
```diff
@@ -7,8 +7,8 @@ trigger:
 
 strategy:
   matrix:
-    python-3.6:
-      imageTag: '3.6'
+    python-3.8:
+      imageTag: '3.8'
```
abuccts, line 11: why it cannot run on python3.6... it may break existing pipelines which reply on ubuntu18 and python3.6

abuccts (changes requested):

.azure-pipelines/cpu-unit-test.yml
cp5555: Shall we just add 3.8? Not replace 3.6?

yukirora: updated to add 3.8  and keep 3.6 and 3.7

yukirora: the dependency is:

- python 3.6 only support numpy < 1.20
- latest numpy requires networkx >= 2.5
- latest networkx and later requires python3.8+
- latest pandas requires numpy >=1.20.3

so currently:
networkx >= 2.5
numpy>=1.19.2
pandas>=1.15

cp5555: Can we check it in setup.py?

yukirora: updated to install networkx,numpy,pandas according to python version

setup.py
```diff
@@ -159,11 +159,16 @@ def run(self):
...
+        'networkx<=2.6;python_version=="3.7"',
+        'networkx>=2.7;python_version>"3.7"',
+        'numpy<1.20.0;python_version=="3.6"',
+        'numpy>=1.20.3;python_version>="3.7"',
         'omegaconf==2.0.6',
         'openpyxl>=3.0.7',
-        'pandas==1.1.5',
+        'pandas<1.2.0;python_version=="3.6"',
+        'pandas<1.4.0;python_version=="3.7"',
+        'pandas>=1.4.0;python_version>"3.7"',
```
abuccts, line 171: > networkx >= 2.5
numpy>=1.19.2
pandas>=1.1.5

will these conditions work directly? setuptools will resolve them automatically as long as there's no conflict?

yukirora: if it's the ideal case that there's a clean env that's ok, but sometimes the lcoal env may pre-installed some packages, and we use --system-site-packages to create venv, so if local env pre-isntalled 1.19.2 numpy while the latest pandas installed when installing superbench, it will cause error

abuccts: > if local env pre-isntalled 1.19.2 numpy while the latest pandas installed

I think this means before installing superbench package the existing python packages already have conflict?
Maybe we can check the return code of [`python -m pip check`](https://pip.pypa.io/en/stable/cli/pip_check/) in the beginning of setup.py? It's hard to cover all possible conflicts (e.g., specify python conditions for all packages) so better to rely on setuptools/pip.

Here's an example output of your proposed case:
```
# pip check
pandas 1.5.3 has requirement numpy>=1.20.3; python_version < "3.10", but you have numpy 1.19.2.
# echo $?
1
```

---

Or can we remove `--system-site-packages` in the document? start from clean env sounds better and can avoid conflicts with apt packages as well.

.azure-pipelines/cpu-unit-test.yml
```diff
@@ -7,10 +7,12 @@ trigger:
...
     python-3.7:
       imageTag: '3.7'
+    python-3.6:
+      imageTag: '3.6'
     # TODO
     #python-latest:
```
abuccts, line 17: keep ascending order? add 3.8 in the end

abuccts (approved):

.azure-pipelines/cpu-unit-test.yml
yukirora: updated

PR #479, opened 2023-02-20
Benchmarks: Revision - Support flexible warmup and non-random data initialization in cublas-benchmark 

yukirora:
**Description**
revise cublas-benchmark for flexible warmup and fill data with fixed number for perf test to improve the running efficiency.

**Major Revision**
- remove num_in_steps for warmup to support more flexible warmup setting for users
- Add support to generate input with fixed number for perf test


superbench/benchmarks/micro_benchmarks/cublas_function/cublas_benchmark.h
```diff
@@ -469,13 +489,12 @@ int CublasFunction::check_result(int batch_count, cuComplex *Result_3_0, std::co
  */
 void CublasFunction::benchmark() {
     // Malloc memory for input and output data
-    this->prepare_tensor();
+    bool random_data = this->correctness ? true : false;
```
cp5555, line 492: I don't think we would like to hard-code this. We only enable to configure it from user and disable random for performance by default

superbench/benchmarks/micro_benchmarks/cublas_function.py
```diff
@@ -190,7 +190,7 @@ def add_parser_arguments(self):
         self._parser.add_argument(
```
cp5555, line 190: Would you please check document as well?

superbench/benchmarks/micro_benchmarks/cublas_function/cublas_benchmark.h
```diff
@@ -228,47 +228,67 @@ class CublasFunction {
...
+        srand(random_seed);
+        for (int i = 0; i < m_ * k_ * batch_count_; i++) {
+            Parameter_0_0_host[i] = ((float)rand() / (float)(RAND_MAX));
+        }
+        for (int i = 0; i < k_ * n_ * batch_count_; ++i) {
+            Parameter_1_0_host[i] = ((float)rand() / (float)(RAND_MAX));
+        }
+    } else {
+        // memset the input data to fixed float value
+        memset(Parameter_0_0_host, 1,
```
cp5555, line 242: For 1, maybe there are some optimizations. Shall we choose another number?

abuccts (approved):
do you need to update cudnn_function as well?

superbench/benchmarks/micro_benchmarks/cublas_function.py
```diff
@@ -190,7 +190,7 @@ def add_parser_arguments(self):
...
             '--num_warmup',
             type=int,
-            default=8,
+            default=8 * 1000,
```
rhewett, line 193: I don't understand this change.  The default warm-up iterations should be small or 0.

superbench/benchmarks/micro_benchmarks/cublas_function/cublas_benchmark.h
```diff
@@ -469,13 +494,12 @@ int CublasFunction::check_result(int batch_count, cuComplex *Result_3_0, std::co
  */
 void CublasFunction::benchmark() {
     // Malloc memory for input and output data
-    this->prepare_tensor();
+    bool random = this->correctness ? true : this->random_data;
```
rhewett, line 497: This looks good to me!

rhewett (commented):
I think the random fix looks good.  I have a few more thoughts here:
* The random initialization can be accelerated by adding -O3 to the host compiler flags.  After some local testing, this cuts about 35% of the run time.
* The random initialization can be further accelerated using OMP.
* The random initialization can be accelerated by moving it to the device.

The issue with the for loops in the warm-up still remains.

cp5555:
> I think the random fix looks good. I have a few more thoughts here:
> 
> * The random initialization can be accelerated by adding -O3 to the host compiler flags.  After some local testing, this cuts about 35% of the run time.
> * The random initialization can be further accelerated using OMP.
> * The random initialization can be accelerated by moving it to the device.
> 
> The issue with the for loops in the warm-up still remains.

Very good comments. I think we can check it and optimize it in next PR

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/cublas_function.py
```diff
@@ -236,6 +236,12 @@ def add_parser_arguments(self):
             required=False,
             help='The acceptable error bound for correctness check.',
         )
+        self._parser.add_argument(
+            '--random_data',
+            action='store_true',
+            default=False,
+            help='Enable random data generation for performance test.',
```
cp5555, line 243: Would you please clarify the default setting in help

yukirora: sure, updated

superbench/benchmarks/micro_benchmarks/cublas_function.py
yukirora: this is to match our initial design for training scenario and to keep it consistent with our previous validation setting.

rhewett (approved):

PR #483, opened 2023-03-01
Pin setuptools version to v65.7.0

abuccts:
Pin setuptools version to [v65.7.0](https://setuptools.pypa.io/en/latest/history.html#v65-7-0) to avoid breaking changes pypa/setuptools#2497 since v66.0.0, see pypa/setuptools#3772 for reference.

### DIFF
--- a/.azure-pipelines/cpu-unit-test.yml
+++ b/.azure-pipelines/cpu-unit-test.yml
@@ -29,7 +29,7 @@ steps:
       echo "##vso[task.prependpath]$HOME/.local/bin"
     displayName: Export path
   - script: |
-      python3 -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


cp5555 (approved):

.azure-pipelines/cpu-unit-test.yml
```diff
@@ -29,7 +29,7 @@ steps:
       echo "##vso[task.prependpath]$HOME/.local/bin"
     displayName: Export path
   - script: |
-      python3 -m pip install --upgrade pip
+      python3 -m pip install --upgrade pip setuptools==65.7
```
cp5555, line 32: Is there any issue if we always keep an old version?

yukirora (approved):

PR #490, opened 2023-03-16
Benchmarks: Bug fix - cublas cannot run all supported configs due to unsupported config's interruption for Hopper GPU

yukirora:
**Description**
fix bug of cublas cannot run all configs due to unsupported config's interruption for Hopper GPU.

**Major Revision**
- identify hopper arch to exclude the unsupported error
- revise micro_base to avoid termination of running the remaining commands run when one command failed

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/micro_base.py
+++ b/superbench/benchmarks/micro_benchmarks/micro_base.py
@@ -166,6 +166,7 @@ class MicroBenchmarkWithInvoke(MicroBenchmark):
         Return:
             True if run benchmark successfully.
         """
+        success = True
         for cmd_idx in range(len(self._commands)):
             logger.info(
                 'Execute command - round: {}, benchmark: {}, command: {}.'.format(
@@ -181,13 +182,13 @@ class MicroBenchmarkWithInvoke(MicroBenchmark):
                         self._curr_run_index, self._name, output.stdout
                     )
                 )
-                return False
+                success = False
             else:
                 if not self._process_raw_result(cmd_idx, output.stdout):
                     self._result.set_return_code(ReturnCode.MICROBENCHMARK_RESULT_PARSING_FAILURE)
-                    return False
+                    success = False
 
-        return True
+        return success
 
     @abstractmethod
     def _process_raw_result(self, cmd_idx, raw_output):
--- a/superbench/benchmarks/micro_benchmarks/cublas_function.py
+++ b/superbench/benchmarks/micro_benchmarks/cublas_function.py
@@ -11,6 +11,7 @@ import statistics
 from superbench.common.utils import logger
 from superbench.benchmarks import Platform, BenchmarkRegistry, ReturnCode
 from superbench.benchmarks.micro_benchmarks import MicroBenchmarkWithInvoke
+from superbench.common.utils import device_manager as dm
 
 
 class CublasBenchmark(MicroBenchmarkWithInvoke):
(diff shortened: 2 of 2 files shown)


superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -166,6 +166,7 @@ def _benchmark(self):
         Return:
             True if run benchmark successfully.
         """
+        success = True
```
cp5555, line 169: It should 'ret'?

superbench/benchmarks/micro_benchmarks/cudnn_function.py
```diff
@@ -440,14 +442,23 @@ def _process_raw_result(self, cmd_idx, raw_output):
...
+            if self._capability == 9.0 and 'CUDNN_STATUS_NOT_SUPPORTED' in raw_output:
+                logger.warning(
+                    'CUDNN_STATUS_NOT_SUPPORTED error in running cudnn test on Hopper GPU - round: {}, \
+                        index of cmd: {},  benchmark: {}, raw data: {}'.format(
+                        self._curr_run_index, cmd_idx, self._name, raw_output
+                    )
                 )
-            )
-            return False
+                self._result.add_result(metric.lower() + '_time', -1)
```
cp5555, line 454: We have return code, why do we need this?

superbench/benchmarks/micro_benchmarks/micro_base.py
```diff
@@ -181,13 +182,13 @@ def _benchmark(self):
...
-                return False
+                success = False
             else:
                 if not self._process_raw_result(cmd_idx, output.stdout):
                     self._result.set_return_code(ReturnCode.MICROBENCHMARK_RESULT_PARSING_FAILURE)
-                    return False
+                    success = False
 
-        return True
+        return success
```
cp5555, line 191: It looks like we don't have return code for each?

superbench/benchmarks/micro_benchmarks/cudnn_function.py
```diff
@@ -440,14 +442,23 @@ def _process_raw_result(self, cmd_idx, raw_output):
                     self._curr_run_index, cmd_idx, self._name, raw_output, str(e)
                 )
             )
-            return False
+            error = True
         if error:
-            logger.error(
-                'Error in running cudnn test - round: {}, index of cmd: {}, benchmark: {}, raw data: {}'.format(
-                    self._curr_run_index, cmd_idx, self._name, raw_output
+            if self._capability == 9.0 and 'CUDNN_STATUS_NOT_SUPPORTED' in raw_output:
```
abuccts, line 447: is it possible to just disable the configs that cannot run?
some configs also fail on A100, and the error code can also be CUDNN_STATUS_NOT_INITIALIZED

yukirora: disable the configs looks weird, I think for this pr, we just make it be able to run even there's one command failed, then it should work for both cases

superbench/benchmarks/micro_benchmarks/micro_base.py
yukirora: updated

superbench/benchmarks/micro_benchmarks/cudnn_function.py
yukirora: since the return code is to indicate the whole benchmark not each command/config, so I added -1 for failed config to keep the metrics

superbench/benchmarks/micro_benchmarks/micro_base.py
yukirora: yes

cp5555 (approved):

PR #491, opened 2023-03-16
Monitor - Support cgroup V2 when read system metrics.

guoshzhao:
**Description**
Since ubuntu 22.04 will use cgroup V2 and the file structure changed. Modify the monitor to adapt to cgroup v1 and v2.


cp5555 (approved):

superbench/monitor/monitor.py
```diff
@@ -80,8 +99,12 @@ def __preprocess(self):
...
                 return False
         else:
-            self._cpu_file = '/sys/fs/cgroup/cpuacct/cpuacct.stat'
-            self._mem_file = '/sys/fs/cgroup/memory/memory.usage_in_bytes'
+            if self.__cgroup == 1:
+                self._cpu_file = '/sys/fs/cgroup/cpuacct/cpuacct.stat'
+                self._mem_file = '/sys/fs/cgroup/memory/memory.usage_in_bytes'
+            else:
+                self._cpu_file = '/sys/fs/cgroup/cpu.stat'
+                self._mem_file = '/sys/fs/cgroup/memory.stat'
```
abuccts, line 107: is this file used? maybe remove it?

abuccts (approved):

superbench/monitor/monitor.py
```diff
@@ -39,6 +39,16 @@ def __init__(self, container_name, sample_duration, sample_interval, output_file
 
         self.__output_handler = open(self.__output_file, 'a')
 
+        self.__cgroup = 1
```
cp5555, line 42: Why cgroup = 1?

guoshzhao: It is the cgroup version, 1 and 2.

superbench/monitor/monitor.py
guoshzhao: Currently not parsed, we can keep the file variable here in case will parse in the future.

superbench/monitor/monitor.py
cp5555: Great!

superbench/monitor/monitor.py
```diff
@@ -215,13 +238,21 @@ def __get_process_cpu_ticks(self):
...
+                        items = line.split()
+                        if items[0] == 'user':
+                            user_time = int(items[1])
+                        elif items[1] == 'system':
+                            system_time = int(items[1])
+                else:
+                    for line in f:
+                        items = line.split()
+                        if items[0] == 'user_usec':
+                            user_time = int(items[1]) / 10000
```
abuccts, line 252: what's the unit of the metric here? usec is 10^-6 sec?

superbench/monitor/monitor.py
```diff
@@ -215,13 +238,21 @@ def __get_process_cpu_ticks(self):
...
+                        elif items[1] == 'system_usec':
+                            system_time = int(items[1]) / 10000
+            return user_time + system_time
```
abuccts, line 255: read usage_usec directly?

superbench/monitor/monitor.py
guoshzhao: Not sure about the unit, but `/10000` can align with the v1's value which can also be got from `/proc/stat`

guoshzhao: int and float will not affect the following calculation.

superbench/monitor/monitor.py
guoshzhao: usage = (end_clock - begin_clock) / duration.
The duration can be controlled by monitor itself.

PR #504, opened 2023-04-04
Analyzer: Fix bug in python3.8 due to pandas api change

yukirora:
**Description**
Analyzer: Fix bug in python3.8 due to pandas api change.

**Major Revision**
- force check numeric only in dataframe for analysis
- dataframe.append -> pd.concat
- pd.ExcelWriter.save() -> pd.ExcelWriter.close()

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -163,7 +163,7 @@ setup(
         'numpy>=1.19.2',
         'omegaconf==2.0.6',
         'openpyxl>=3.0.7',
-        'pandas>=1.1.5, <2.0.0',
+        'pandas>=1.1.5',
         'pssh @ git+https://github.com/lilydjwg/pssh.git@v2.3.4',
         'pyyaml>=5.3',
         'requests>=2.27.1',
--- a/setup.py
+++ b/setup.py
@@ -163,7 +163,7 @@ setup(
         'numpy>=1.19.2',
         'omegaconf==2.0.6',
         'openpyxl>=3.0.7',
-        'pandas>=1.1.5',
+        'pandas>=1.1.5, <2.0.0',
         'pssh @ git+https://github.com/lilydjwg/pssh.git@v2.3.4',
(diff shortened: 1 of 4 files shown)


abuccts (commented):
maybe fix the pandas to specific version? they're breaking changes starting from [2.0.0](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#backwards-incompatible-api-changes)

yukirora:
> maybe fix the pandas to specific version? they're breaking changes starting from [2.0.0](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#backwards-incompatible-api-changes)

sure, updated to fix version to <2.0.0

superbench/analyzer/result_summary.py
```diff
@@ -217,7 +217,7 @@ def output_summary_in_excel(self, raw_data_df, summary, output_path):
                 file_handler.merge_column_in_excel(worksheet, row, 1)
             else:
                 logger.error('ResultSummary: excel_data_output - summary is empty.')
-            writer.save()
+            writer.close()
```
cp5555, line 220: So, it will save the excel when closing?

setup.py
```diff
@@ -163,7 +163,7 @@ def run(self):
         'numpy>=1.19.2',
         'omegaconf==2.0.6',
         'openpyxl>=3.0.7',
-        'pandas>=1.1.5',
+        'pandas>=1.1.5, <2.0.0',
```
cp5555, line 166: Why not support >2.0.0 directly?

superbench/analyzer/result_summary.py
yukirora: yes, it's synonym for save, they replace save to close in latest version to make it more file-like

setup.py
yukirora: updated

PR #512, opened 2023-04-11
Remove unreachable condition when write host list

abuccts:
Remove unreachable condition when write host list.


cp5555 (approved):

guoshzhao:
What is this change used for? Let `torch.distributed` mode support different pattern?

RyoYang (approved):

guoshzhao (approved):

PR #522, opened 2023-04-17
ModelBenchmarks - Fix early stop logic due to num_steps.

guoshzhao:
**Description**
Model benchmarks can stop due to `num_steps` or `duration` config which will take effect when the value is set greater than 0.
If both are set greater than 0, the earliest condition reached will work.

### DIFF
--- a/docs/superbench-config.mdx
+++ b/docs/superbench-config.mdx
@@ -347,7 +347,7 @@ For Model-Benchmark, there have some parameters that can control the elapsed tim
 * num_warmup: the number of warmup step.
 * num_steps: the number of test step.
 
-If `duration > 0` and `num_warmup + num_steps > 0`, then benchmark will take the least as the elapsed time. Otherwise only one of them will take effect.
+If `duration > 0` and `num_steps > 0`, then benchmark will take the least as the elapsed time. Otherwise only one of them will take effect.
 
 ## `Mode` Schema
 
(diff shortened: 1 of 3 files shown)


superbench/benchmarks/model_benchmarks/model_base.py
```diff
@@ -374,7 +374,7 @@ def _is_finished(self, curr_step, curr_time):
 
         if (
             (self._args.duration > 0 and (curr_time - self._sub_benchmark_start_time) >= self._args.duration)
-            or (total_steps > 0 and curr_step >= total_steps)
+            or (self._args.num_steps > 0 and curr_step >= total_steps)
```
yukirora, line 377: the previous design is when 'num_warmup + num_steps < 0' and duration>0 it only use the duration, now you want to change to when 'num_step < ' and duration >0, it will only use the duration for convenience, if so, you need to change the doc accordingly

guoshzhao: Yes, it is more clear to only use num_step in case users will use a big value for num_warmup to affect the value of num_warmup + num_steps.
Thanks, I will revise the doc.

guoshzhao: Done.

cp5555 (approved):

docs/superbench-config.mdx
```diff
@@ -347,7 +347,7 @@ For Model-Benchmark, there have some parameters that can control the elapsed tim
 * num_warmup: the number of warmup step.
```
cp5555, line 347: Would you please add `num_warmup should be positive integer`?

yukirora (approved):

PR #524, opened 2023-04-18
Benchmarks - Revise step time collection in distributed inference benchmark

yzygitzh:
**Description**
This commit revises distributed inference benchmark to give a unified step time result by taking maximum step times of different GPUs.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/dist_inference.py
+++ b/superbench/benchmarks/micro_benchmarks/dist_inference.py
@@ -14,6 +14,7 @@ from superbench.common.utils import logger
 from superbench.benchmarks import DistributedImpl, DistributedBackend, BenchmarkRegistry, ReturnCode, Precision
 from superbench.benchmarks.micro_benchmarks import MicroBenchmark
 from superbench.benchmarks.context import Enum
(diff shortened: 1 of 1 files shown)


guoshzhao (approved):

superbench/benchmarks/micro_benchmarks/dist_inference.py
```diff
@@ -390,7 +391,7 @@ def _process_data(self, step_times):
         Return:
             True if _process_data succeeds.
         """
-        if not self._process_numeric_result('step_times', step_times, cal_percentile=True):
+        if not self._process_numeric_result('step_times', step_times, reduce_type=ReduceType.MAX, cal_percentile=True):
```
abuccts, line 394: this is max on single gpu results? do you need `torch.dist.all_reduce(steptime, op=torch.dist.ReduceOp.MAX)` to get max result across gpus?

yzygitzh: Checked with @guoshzhao and @yukirora and the experiment results, it's max across ranks.

abuccts (approved):

PR #530, opened 2023-05-08
Runner - Add signal handler in runner

abuccts:
Add signal handler in runner to gracefully exit when receiving SIGINT (<kbd>Ctrl</kbd>+<kbd>C</kbd>) or SIGTERM during benchmark execution.

### DIFF
--- a/superbench/runner/runner.py
+++ b/superbench/runner/runner.py
@@ -242,9 +242,10 @@ class SuperBenchRunner():
             signum (int): Signal number.
             frame (FrameType): Timeout frame.
         """
-        logger.info('Killed by %s, exiting ...', signal.Signals(signum).name)
-        self.cleanup()
-        sys.exit(128 + signum)
+        if signum == signal.SIGINT or signum == signal.SIGTERM:
+            logger.info('Killed by %s, exiting ...', signal.Signals(signum).name)
+            self.cleanup()
+            sys.exit(128 + signum)
 
     def __create_results_summary(self):    # pragma: no cover
         """Create the result summary file of all nodes."""
--- a/superbench/runner/runner.py
+++ b/superbench/runner/runner.py
@@ -4,8 +4,10 @@
 """SuperBench Runner."""
 
 import os
+import sys
 import json
 import random
+import signal
 from pathlib import Path
 from pprint import pformat
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


superbench/runner/runner.py
```diff
@@ -233,6 +235,17 @@ def fetch_results(self):    # pragma: no cover
...
         )
 
+    def __signal_handler(self, signum, frame):
+        """Signal handler for runner.
+
+        Args:
+            signum (int): Signal number.
+            frame (FrameType): Timeout frame.
+        """
+        logger.info('Killed by %s, exiting ...', signal.Signals(signum).name)
```
cp5555, line 245: Do we need to check signum?

abuccts: only SIGINT and SIGTERM are trapped
https://github.com/microsoft/superbenchmark/blob/9a6dc23fed01896f4e265f2b677aa400b57be7b8/superbench/runner/runner.py#L463-L464

cp5555: I mean whether we should check here. For example, someone change the code in runner but forget here. It will trigger the non-expected issue.

abuccts: sure, updated

cp5555 (approved):

superbench/runner/runner.py
```diff
@@ -233,6 +235,18 @@ def fetch_results(self):    # pragma: no cover
...
         )
 
+    def __signal_handler(self, signum, frame):
+        """Signal handler for runner.
+
+        Args:
+            signum (int): Signal number.
+            frame (FrameType): Timeout frame.
+        """
+        if signum == signal.SIGINT or signum == signal.SIGTERM:
```
guoshzhao, line 245: I remember that there also has one signal named `SIGKILL`, do we need to catch it?

superbench/runner/runner.py
```diff
@@ -438,12 +452,17 @@ def _run_proc(self, benchmark_name, mode, vars):
             # we do not expect timeout in ansible unless subprocess hangs
             ansible_runner_config['timeout'] = timeout + 60
 
-        rc = self._ansible_client.run(ansible_runner_config, sudo=(not self._docker_config.skip))
+        # overwrite ansible runner's default signal handler with main process's
```
guoshzhao, line 455: Can't overwrite ansible runner's default signal handler using your handler?

superbench/runner/runner.py
cp5555: It looks like we cannot catch it from https://docs.python.org/3/library/signal.html

guoshzhao (approved):

superbench/runner/runner.py
abuccts: ansible is running in child process, current handler is registered in main process and traps signals in both ansible runner and other python logic

PR #536, opened 2023-06-01
Benchmarks - Add support for DirectX GPU platform

yukirora:
**Description**
Add support for DirectX GPU platform.

**Major Revision**
- Add DirectX platform for benchmark registry
- Add gpu_vendor identify for AMD and NVIDIA with win driver

### DIFF
--- a/superbench/benchmarks/__init__.py
+++ b/superbench/benchmarks/__init__.py
@@ -16,7 +16,7 @@ BenchmarkRegistry = LazyImport(
         map(
             importlib.import_module, [
                 'superbench.benchmarks.{}'.format(module)
-                for module in ['model_benchmarks', 'micro_benchmarks', 'docker_benchmarks']
+                for module in ['model_benchmarks', 'micro_benchmarks', 'docker_benchmarks', 'directx_benchmarks']
             ]
         )
     )
--- a/superbench/benchmarks/__init__.py
+++ b/superbench/benchmarks/__init__.py
@@ -16,7 +16,7 @@ BenchmarkRegistry = LazyImport(
         map(
             importlib.import_module, [
                 'superbench.benchmarks.{}'.format(module)
-                for module in ['model_benchmarks', 'micro_benchmarks', 'docker_benchmarks', 'directx_benchmarks']
+                for module in ['model_benchmarks', 'micro_benchmarks', 'docker_benchmarks']
             ]
         )
     )
--- a/superbench/common/devices/gpu.py
+++ b/superbench/common/devices/gpu.py
@@ -29,6 +29,10 @@ class GPU():
             if not list(Path('/dev/dri').glob('card*')):
                 logger.warning('Cannot find AMD GPU device.')
             return 'amd'
+        if list(Path(r'C:\Windows\System32').glob('*DriverStore/FileRepository/u*.inf_amd64_*/*/aticfx64.dll')):
+            return 'amd-win'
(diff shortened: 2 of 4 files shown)


superbench/benchmarks/__init__.py
```diff
@@ -16,7 +16,7 @@
         map(
             importlib.import_module, [
                 'superbench.benchmarks.{}'.format(module)
-                for module in ['model_benchmarks', 'micro_benchmarks', 'docker_benchmarks']
+                for module in ['model_benchmarks', 'micro_benchmarks', 'docker_benchmarks', 'directx_benchmarks']
```
cp5555, line 19: I think it's different with this. model, micro, and docker benchmark should be orthogonal with DirectX benchmark. DirectX should be similar as CPU, CUDA and ROCm

superbench/common/devices/gpu.py
```diff
@@ -29,6 +29,10 @@ def get_vendor(self):
             if not list(Path('/dev/dri').glob('card*')):
                 logger.warning('Cannot find AMD GPU device.')
             return 'amd'
+        if list(Path(r'C:\Windows\System32').glob('*DriverStore/FileRepository/u*.inf_amd64_*/*/aticfx64.dll')):
```
cp5555, line 32: is it too long?

superbench/executor/executor.py
```diff
@@ -87,6 +87,10 @@ def __get_platform(self):
                 return Platform.CUDA
             elif gpu.vendor == 'amd':
                 return Platform.ROCM
+            elif gpu.vendor == 'amd-win':
+                return Platform.DIRECTX
+            elif gpu.vendor == 'nvidia-win':
+                return Platform.DIRECTX
```
cp5555, line 93: Do we need to separate AMD or Nvidia? If so, we should separate them here as well. Otherwise, why do we keep amd-win and nvidia-win?

superbench/common/devices/gpu.py
yukirora: I think no, it can pass lint

superbench/benchmarks/__init__.py
yukirora: do you think the source code of all DirectX benchmarks should be under microbenchmarks directly, or microbenchmarks/directxbenchmark?

superbench/executor/executor.py
yukirora: separated

superbench/common/devices/gpu.py
```diff
@@ -29,6 +29,10 @@ def get_vendor(self):
...
+            return 'amd-win'
+        if list(Path(r'C:\Windows\System32').glob('*DriverStore/FileRepository/nv*.inf_amd64_*/nvapi64.dll')):
+            return 'nvidia-win'
```
guoshzhao, line 35: Should we first judge nvidia, then amd for consistence, just like Linux platform?

superbench/executor/executor.py
```diff
@@ -87,6 +87,10 @@ def __get_platform(self):
                 return Platform.CUDA
             elif gpu.vendor == 'amd':
                 return Platform.ROCM
+            elif gpu.vendor == 'amd-win':
+                return Platform.DIRECTXAMD
+            elif gpu.vendor == 'nvidia-win':
+                return Platform.DIRECTXNVIDIA
```
guoshzhao, line 93: Same as last comment.

guoshzhao (approved):

abuccts (approved):

cp5555 (approved):

PR #538, opened 2023-06-13
Update error message in setup

abuccts:
Update error message in setup, require wheel for pip>=23.1.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,10 @@ import superbench
 try:
     pkg_resources.require(['pip>=18', 'setuptools>=45, <66'])
 except (pkg_resources.VersionConflict, pkg_resources.DistributionNotFound):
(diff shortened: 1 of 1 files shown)


RyoYang (approved):

setup.py
```diff
@@ -20,7 +20,10 @@
 try:
     pkg_resources.require(['pip>=18', 'setuptools>=45, <66'])
 except (pkg_resources.VersionConflict, pkg_resources.DistributionNotFound):
-    print('Try update pip/setuptools versions, for example, python3 -m pip install --upgrade pip setuptools==65.7')
+    print(
+        '\033[93mTry update pip/setuptools versions, for example, '
```
yzygitzh, line 24: Shall we make \033[93m things standalone variables? It's hard to understand.

yzygitzh (approved):

setup.py
abuccts: it's only used once here and is a standard asci code

PR #541, opened 2023-06-29
Benchmarks - Update result parsing in tensorrt inference

abuccts:
* Update result parsing for newer tensorrt versions
* Update arguments when load torchvision models


superbench/benchmarks/micro_benchmarks/_export_torch_to_onnx.py
```diff
@@ -129,7 +130,8 @@ def export_torchvision_model(self, model_name, batch_size=1):
         if not self.check_torchvision_model(model_name):
             return ''
         file_name = str(self._onnx_model_path / (model_name + '.onnx'))
-        model = getattr(torchvision.models, model_name)(pretrained=False).eval().cuda()
+        args = {'pretrained': False} if version.parse(torchvision.__version__) < version.parse('0.13') else {}
```
guoshzhao, line 133: Could we add more comments about the hardcode version here to avoid confusion in the future?

guoshzhao (approved):

superbench/benchmarks/micro_benchmarks/_export_torch_to_onnx.py
abuccts: added

cp5555 (approved):

PR #543, opened 2023-06-29
Benchmarks: Build Pipeline - add AMF in third party and build AMF encoding latency test

yukirora:
**Description**
add AMF in third party and build AMF encoding latency test.


cp5555 (approved):

.gitmodules
```diff
@@ -21,3 +21,6 @@
 [submodule "third_party/gpu-burn"]
 	path = third_party/gpu-burn
 	url = https://github.com/wilicc/gpu-burn.git
+[submodule "third_party/AMF"]
+	path = third_party/AMF
+	url = https://github.com/GPUOpen-LibrariesAndSDKs/AMF.git
```
abuccts, line 26: the module seems too large (GB level) and is not used in most cases
maybe move to dockerfile or makefile and clone only when building directx?

yukirora: moved to the third_party/Makefile directly

.github/workflows/build-image.yml
```diff
@@ -64,6 +64,17 @@ jobs:
           sudo apt-get clean
           sudo docker rmi $(sudo docker images --format "{{.Repository}}:{{.Tag}}" --filter=reference="node" --filter=reference="buildpack-deps")
           df -h
+      - name: Free Up GitHub Actions Ubuntu Runner Disk Space 🔧
```
abuccts, line 67: don't need any more?

yukirora: ok, removed

abuccts (approved):

PR #545, opened 2023-06-30
CI/CD - Support DirectX test pipeline

yukirora:
**Description**
Support DirectX test pipeline.


cp5555 (approved):

.github/workflows/directx-unit-test.yml
```diff
@@ -0,0 +1,65 @@
...
+      shell: pwsh
+    - name: Build benchmarks
+      run: |
+        $env:SB_MICRO_PATH=$PWD
+        make directxbuild
+      shell: pwsh
+    - name: Build Third Party
+      run: |
+        $env:SB_MICRO_PATH=$PWD
+        make -C third_party directx_amd
```
abuccts, line 54: do you need to build in every run?

setup.py
```diff
@@ -97,19 +97,24 @@ class Tester(Command):
...
         """Set final values for options that this command supports."""
-        pass
+        if self.type not in {None, 'directx', 'cuda', 'rocm'}:    # replace with your test types
+            raise Exception(f'Invalid type: {self.type}')
 
     def run(self):
         """Run pytest."""
-        errno = os.system('python3 -m pytest -v --cov=superbench --cov-report=xml --cov-report=term-missing tests/')
+        command = 'python3 -m pytest -v --cov=superbench --cov-report=xml --cov-report=term-missing tests/'
+        command += f' -k test_{self.type}' if self.type else ''
```
abuccts, line 116: why this is needed if you set env to skip test?

abuccts (commented):
can you fix the unit test first?

setup.py
yukirora: because I need to skip other benchmarks unit test, they can not successfully run on win, and I can not set a skipunittest for all other benchmarks

.github/workflows/directx-unit-test.yml
yukirora: if I do not build it then what if there's no bin for unit test?  I think we need to build here and control not build duplicate in makefile

yukirora:
> can you fix the unit test first?

that's because the depency on #543 PR, if that PR merged, then this PR's UT will pass

.github/workflows/directx-unit-test.yml
abuccts: can you run the unit test in container? maybe just add a "docker run superbench/main:win2004 python setup.py test" task after docker build?

setup.py
yukirora: not only ansible, I have tried, there're too many issues, some are permission issues when create file, some are bin issues, etc.. I even can hardly check all issues in short time

yukirora: updated to use pytest command + '-k' in pipeline file directly

.github/workflows/build-win.yml
```diff
@@ -44,3 +44,26 @@ jobs:
         TAG: superbench/main:win2004
         USER: ${{ secrets.DOCKERHUB_USERNAME }}
         PASS: ${{ secrets.DOCKERHUB_TOKEN }}
+  directx-unit-test:
+    name: DirectX unit test
+    needs: docker
+    runs-on: [self-hosted, windows, x64, win2004]
+    steps:
```
abuccts, line 51: remove? they should be in the same job to reuse latest built image

.github/workflows/build-win.yml
```diff
@@ -44,3 +44,26 @@ jobs:
...
+    - name: Run unit tests
+      run: |
+        docker run --isolation process \
+        -e SB_MICRO_PATH="$PWD" \
```
abuccts, line 55:
```
`
```

.github/workflows/build-win.yml
```diff
@@ -44,3 +44,26 @@ jobs:
...
+        -e SB_TEST_ROCM="0" \
+        -e SB_TEST_PYTORCH="0" \
+        -e SB_TEST_DIRECTX="1" \
+        --device class/5B45201D-F2F2-4F3B-85BB-30FF1F953599 \
+        superbench/main:win2004 \
+        python3 -m pytest -v --cov=superbench --cov-report=xml --cov-report=term-missing tests/ -k test_directx
+      shell: pwsh
+    - name: Report coverage results
+      run: |
+        curl -s https://codecov.io/bash | bash -s -- -cF directx-unit-test
```
abuccts, line 66: check https://docs.codecov.com/docs/testing-with-docker#codecov-inside-docker

.github/workflows/build-win.yml
yukirora: why need to be in the same job, I think it's reuse the latest build image now since I have defined the dependency of these 2 jobs?

.github/workflows/build-win.yml
yukirora: ok, updated

.github/workflows/build-win.yml
yukirora: thanks, updated, pls help check

tests/runner/test_ansible.py
```diff
@@ -10,8 +10,6 @@
 
 from omegaconf import OmegaConf
 
-from superbench.runner.ansible import AnsibleClient
```
abuccts, line 13: use lazyimport?
```
AnsibleClient = LazyImport('superbench.runner.ansible', 'AnsibleClient')
```

.github/workflows/build-win.yml
```diff
@@ -44,3 +44,27 @@ jobs:
...
+        echo "$env:PATH;C:\Program Files\Git\bin" | Out-File -FilePath $env:GITHUB_PATH -Append -Encoding utf8
+    - name: Bash to get codecov env
+      run: |
+        ci_env=`bash <(curl -s https://codecov.io/env)`
+        echo "ci_env=$ci_env" >> $GITHUB_ENV
+      shell: bash
+    - name: Run unit tests inside docker
+      run: |
+        $command="curl -s -L https://uploader.codecov.io/latest/windows/codecov.exe -o codecov.exe && python -m pip install .[test] && python -m pytest -v --cov=superbench --cov-report=xml --cov-report=term-missing tests/ -k test_directx && codecov -cF directx-unit-test"
+        docker run `
```
abuccts, line 64: add "--rm"?

yukirora: thanks, updated

abuccts (approved):

PR #546, opened 2023-07-03
Benchmarks: micro benchmarks - add python code for DirectXGPUCopy

yukirora:
**Description**
add python code for DirectXGPUCopy.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/directx_gpu_copy_performance.py
+++ b/superbench/benchmarks/micro_benchmarks/directx_gpu_copy_performance.py
@@ -44,7 +44,7 @@ class DirectXGPUCopyBw(MemBwBenchmark):
             '--num_loops',
             type=int,
             required=False,
-            default=100000,
+            default=1000,
             help='Number of copy times to run.',
         )
         self._parser.add_argument(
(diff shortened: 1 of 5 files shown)


superbench/benchmarks/micro_benchmarks/directx_gpu_copy_performance.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, name, parameters=''):
+        """Constructor.
+
+        Args:
+            name (str): benchmark name.
+            parameters (str): benchmark parameters.
+        """
+        super().__init__(name, parameters)
+        self._mem_types = ['htod', 'dtoh']
+        self._bin_name = 'GPUCopyBw.exe'
```
cp5555, line 23: Also the same question as Ziyue: The name is not related to this benchmark.

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/directx_gpu_copy_performance.py
```diff
@@ -0,0 +1,140 @@
...
+            type=int,
+            required=False,
+            default=20,
+            help='Number of warm up copy times to run.',
+        )
+        self._parser.add_argument(
+            '--num_loops',
+            type=int,
+            required=False,
+            default=100000,
```
yzygitzh, line 47: Why do we need such huge num_loops? For gpu_copy I think it's around 1k.

yzygitzh (approved):

superbench/benchmarks/micro_benchmarks/directx_gpu_copy_performance.py
yukirora: renamed to DirectXGPUCopyBw.exe

superbench/benchmarks/micro_benchmarks/directx_gpu_copy_performance.py
yukirora: updated

PR #580, opened 2023-11-23
Monitor - Add support for AMD GPU.

guoshzhao:
**Description**
Add AMD support in monitor.

**Major Revision**
- Add library pyrsmi to collect metrics.
- Currently can get device_utilization, device_power, device_used_memory and device_total_memory.


yukirora:
will you change the exec function in executor.py, no way to enable monitor on rocm platform currently
```
if self.__get_platform() == Platform.CUDA:
                    monitor = Monitor(
                        None, int(self._sb_monitor_config.sample_duration or 10),
                        int(self._sb_monitor_config.sample_interval or 1), self.__get_monitor_path(benchmark_name)
                    )
                    monitor.start()
                else:
                    logger.warning('Monitor can not support ROCM/CPU platform.')
```

cp5555 (approved):

guoshzhao:
> will you change the exec function in executor.py, no way to enable monitor on rocm platform currently
> 
> ```
> if self.__get_platform() == Platform.CUDA:
>                     monitor = Monitor(
>                         None, int(self._sb_monitor_config.sample_duration or 10),
>                         int(self._sb_monitor_config.sample_interval or 1), self.__get_monitor_path(benchmark_name)
>                     )
>                     monitor.start()
>                 else:
>                     logger.warning('Monitor can not support ROCM/CPU platform.')
> ```

Good catch. Thanks.

superbench/config/default.yaml
```diff
@@ -3,7 +3,7 @@ version: v0.9
 superbench:
   enable: null
   monitor:
-    enable: true
+    enable: false
```
cp5555, line 6: Why change this?

guoshzhao: Since lot's of UT will use this default config file and create monitors even don't meant to test it. Just want to accelate the UT process.
I have reverted this change.,

yukirora (approved):

PR #583, opened 2023-12-04
Benchmarks: Micro benchmark - Add graph mode in NCCL/RCCL benchmarks for latency metrics

yzygitzh:
**Description**
Revise NCCL/RCCL benchmarks to add latency metrics.

### DIFF
--- a/tests/benchmarks/micro_benchmarks/test_cuda_nccl_bw_performance.py
+++ b/tests/benchmarks/micro_benchmarks/test_cuda_nccl_bw_performance.py
@@ -74,7 +74,7 @@ class CudaNcclBwBenchmarkTest(BenchmarkTestCase, unittest.TestCase):
         ]
 
         command = bin_names[0] + benchmark._commands[0].split(bin_names[0])[1]
-        expected_command = '{} -b 8 -e 8G -f 2 -g 8 -c 0 -n 20 -w 5'.format(bin_names[0])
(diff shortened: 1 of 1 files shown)


yukirora (approved):
seems this title does not match the content?

superbench/config/azure_ndmv4.yaml
```diff
@@ -73,6 +73,17 @@ superbench:
             NCCL_IB_DISABLE: '0'
       parameters:
         ngpus: 8
+    nccl-lat:default:
```
cp5555, line 76: Why do we change the default?

superbench/benchmarks/micro_benchmarks/cuda_nccl_bw_performance.py
```diff
@@ -88,6 +88,12 @@ def add_parser_arguments(self):
             default=5,
             help='Number of warmup iterations. Default: 5.',
         )
+        self._parser.add_argument(
+            '--graph_iters',
+            type=int,
+            default=0,
+            help='Number of graph launch iterations. Set to 0 to disable graph mode. Default: 0.',
```
cp5555, line 95: Also it's not related to latency as well

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/cuda_nccl_bw_performance.py
yzygitzh: Fixed in PR title

PR #592, opened 2023-12-12
Bug Fix - Fix NUMA Domains Swap Issue in NDv4 Topology File

RyoYang:
**Description**

* Snyc ndv4-topo.xml update with [azhpcimage repo](https://github.com/Azure/azhpc-images/blob/master/topology/ndv4-topo.xml) to fix numa domains swap issue


guoshzhao:
@abuccts Currently only cuda11.1.1 image has this file, do we need to add to other images if user upgrade their images?

cp5555 (approved):

abuccts (approved):

abuccts:
> @abuccts Currently only cuda11.1.1 image has this file, do we need to add to other images if user upgrade their images?

It should be added in all cuda dockerfiles? rocm does not need this currently

guoshzhao (approved):

PR #594, opened 2023-12-14
Benchmarks: Microbenchmark - Improve hipblasLt data type recognition

yzygitzh:
**Description**
Automate hipblasLt data type recognition using cmake symbol looking up feature.


yukirora:
this seems not work, 
image

yzygitzh:
Tried several ways and cannot get it work.. Let's keep current implementation.

PR #595, opened 2023-12-14
Benchmarks: Microbenchmark - Add data type option for NCCL and RCCL tests

yzygitzh:
**Description**
Add data type option for nccl tests.


yukirora (approved):

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/cuda_nccl_bw_performance.py
```diff
@@ -99,6 +99,12 @@ def add_parser_arguments(self):
             action='store_true',
```
cp5555, line 99: Why do we add feature only for NCCL?

yzygitzh: Actually it’s for both NCCL and RCCL.

PR #605, opened 2023-12-25
Benchmarks: Micro benchmarks - add fp8 and initialization for hipblaslt benchmark

yukirora:
**Description**
Benchmarks: Micro benchmarks - add fp8 and initialization for hipblaslt benchmark.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/hipblaslt_function.py
+++ b/superbench/benchmarks/micro_benchmarks/hipblaslt_function.py
@@ -124,9 +124,7 @@ class HipBlasLtBenchmark(BlasLtBaseBenchmark):
             fields = lines[index + 1].strip().split(',')
 
             # Check the number of fields and the format of the first two fields
-            if len(fields) != 23 or not all(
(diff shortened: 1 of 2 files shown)


superbench/benchmarks/micro_benchmarks/hipblaslt_function.py
```diff
@@ -97,9 +124,7 @@ def _process_raw_result(self, cmd_idx, raw_output):
             fields = lines[index + 1].strip().split(',')
 
             # Check the number of fields and the format of the first two fields
```
yzygitzh, line 126: Why we don't check format of first two fields anymore?

yzygitzh (approved):

cp5555 (approved):

superbench/benchmarks/micro_benchmarks/hipblaslt_function.py
yukirora: since I found the fp8 output is 1.xxe6, so my original check is incorrect

PR #614, opened 2024-03-18
Bug Fix - Bug fix for cuda 12.2 dockerfile LD_LIBRARY_PATH issue

RyoYang:
**Description**
Cuda 12.2 image will report undfined symbol error due to incomplete LD_LIBRARY_PATH:

image

### How to reproduce:
1. Deploy sb with cuda12.2 image
```
sb deploy -f local.ini -i superbench/superbench:v0.10.0-cuda12.2
```
2. Enter to the container
```
sudo docker exec -it sb-workspace bash
```
3. Execute `mpirun`:
```
root@sb-container:~# mpirun
mpirun: symbol lookup error: mpirun: undefined symbol: opal_libevent2022_event_base_loop
```
### Fix to fix
* Append hpcx_load into /etc/bash.bashrc for updaing env LD_LIBRARY_PATH in each time

### DIFF
--- a/dockerfile/cuda11.1.1.dockerfile
+++ b/dockerfile/cuda11.1.1.dockerfile
@@ -118,7 +118,7 @@ RUN cd /tmp && \
     rm -rf ./Linux mlc.tgz
 
 ENV PATH="${PATH}" \
-    LD_LIBRARY_PATH="/usr/local/lib:/opt/hpcx/ompi/lib:${LD_LIBRARY_PATH}" \
+    LD_LIBRARY_PATH="/usr/local/lib:${LD_LIBRARY_PATH}" \
     SB_HOME=/opt/superbench \
     SB_MICRO_PATH=/opt/superbench \
     ANSIBLE_DEPRECATION_WARNINGS=FALSE \
--- a/dockerfile/cuda11.1.1.dockerfile
+++ b/dockerfile/cuda11.1.1.dockerfile
@@ -118,7 +118,7 @@ RUN cd /tmp && \
     rm -rf ./Linux mlc.tgz
 
 ENV PATH="${PATH}" \
-    LD_LIBRARY_PATH="/usr/local/lib:${LD_LIBRARY_PATH}" \
+    LD_LIBRARY_PATH="/usr/local/lib:/opt/hpcx/ompi/lib:${LD_LIBRARY_PATH}" \
     SB_HOME=/opt/superbench \
     SB_MICRO_PATH=/opt/superbench \
     ANSIBLE_DEPRECATION_WARNINGS=FALSE \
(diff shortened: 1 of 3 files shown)


dockerfile/cuda11.1.1.dockerfile
```diff
@@ -118,7 +118,7 @@ RUN cd /tmp && \
     rm -rf ./Linux mlc.tgz
 
 ENV PATH="${PATH}" \
-    LD_LIBRARY_PATH="/usr/local/lib:${LD_LIBRARY_PATH}" \
+    LD_LIBRARY_PATH="/usr/local/lib:/opt/hpcx/ompi/lib:${LD_LIBRARY_PATH}" \
```
guoshzhao, line 121: Not sure if we need to add more, such as `/opt/hpcx/ucx/lib`?

RyoYang: Good point and updated!

dockerfile/cuda12.2.dockerfile
```diff
@@ -122,7 +122,15 @@ RUN cd /tmp && \
     rm -rf /tmp/nccl
 
 ENV PATH="${PATH}" \
-    LD_LIBRARY_PATH="/usr/local/lib:${LD_LIBRARY_PATH}" \
+    LD_LIBRARY_PATH="/usr/local/lib:\
+/opt/hpcx/nccl_rdma_sharp_plugin/lib:\
```
abuccts, line 126: is it possible to append `source /opt/hpcx/hpcx-init.sh && hpcx_load` to `~/.bashrc` when installing hpcx? paths here may be changed in future hpcx versions

dockerfile/cuda11.1.1.dockerfile
abuccts, line 121: is this needed, this nvcr base image may use openmpi by default?

RyoYang: It seems was Introduced unexpectedly, removed.

dockerfile/cuda12.2.dockerfile
RyoYang: Good point. However, each time docker run is executed, it mounts a different workspace to /root using -v {{ workspace }}:/root, which deletes all the original files in /root. So, changed to add this file to /opt/superbench at first, and restore the files during docker exec. What do you think of this?

abuccts: I forgot about it. how about appending to system wide `/etc/bash.bashrc`? [it's loaded before `~/.bashrc`](https://askubuntu.com/a/815083/1776723)

RyoYang: Make sense to me and verified.

abuccts (approved):

guoshzhao (approved):

PR #631, opened 2024-07-11
Update omegaconf version to 2.3.0

RyoYang:
Update `omegaconf` version to [2.3.0](https://pypi.org/project/omegaconf/2.3.0/) as omegaconf 2.0.6 has a non-standard dependency specifier PyYAML>=5.1.*. pip 24.1 will enforce this behaviour change.
Discussion can be found at https://github.com/pypa/pip/issues/12063

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -198,7 +198,7 @@ setup(
                 'pydocstyle>=5.1.1',
                 'pytest-cov>=2.11.1',
                 'pytest-subtests>=0.4.0',
-                'pytest>=6.2.2',
+                'pytest>=6.2.2, <=7.4.4',
                 'types-markdown',
                 'types-pkg_resources',
                 'types-pyyaml',
--- a/setup.py
+++ b/setup.py
@@ -164,7 +164,7 @@ setup(
         'natsort>=7.1.1',
         'networkx>=2.5',
         'numpy>=1.19.2',
-        'omegaconf==2.0.6',
+        'omegaconf==2.3.0',
         'openpyxl>=3.0.7',
         'packaging>=21.0',
         'pandas>=1.1.5',
--- a/superbench/runner/runner.py
+++ b/superbench/runner/runner.py
@@ -105,12 +105,12 @@ class SuperBenchRunner():
         Return:
             list: List of benchmarks which will be executed.
         """
-        if self._sb_config.superbench.enable:
+        if 'enable' in self._sb_config.superbench and self._sb_config.superbench.enable:
             if isinstance(self._sb_config.superbench.enable, str):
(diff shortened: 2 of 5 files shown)


superbench/runner/runner.py
```diff
@@ -441,11 +441,11 @@ def _run_proc(self, benchmark_name, mode, vars):
             int: Process return code.
         """
         mode.update(vars)
-        if mode.name == 'mpi' and mode.pattern:
+        if mode.name == 'mpi' and 'pattern' in mode:
```
guoshzhao, line 444: mode.name always exists?

guoshzhao (approved):

superbench/runner/runner.py
RyoYang: Yes,  according to the schema of mode here: https://github.com/microsoft/superbenchmark/blob/main/docs/superbench-config.mdx#:~:text=parallel%3A%20yes-,name,-Mode%20name%20to

cp5555 (approved):

setup.py
```diff
@@ -198,7 +198,7 @@ def run(self):
                 'pydocstyle>=5.1.1',
                 'pytest-cov>=2.11.1',
                 'pytest-subtests>=0.4.0',
-                'pytest>=6.2.2',
+                'pytest>=6.2.2, <=7.4.4',
```
cp5555, line 201: It's strange to set the upper version.

cp5555 (changes requested):
Strange to set the upper version

setup.py
RyoYang: Pytest version 8.2.2 introduced an issue (https://github.com/pytest-dev/pytest/issues/12289) that broke Knack (Issue: https://github.com/microsoft/knack/issues/287). As a result, the pipeline job for Python 3.8 fails due to this issue:
image
I have contacted the Knack team to address this issue. Limiting the pytest version to 7.4.4 (align with python3.7) is a temporary workaround to ensure the SB pipeline passes until Knack provides a fix.

yukirora:
Is this change compatible with the old pip like 22.0.1?

cp5555 (approved):

RyoYang:
> Is this change compatible with the old pip like 22.0.1?
It should be compatible, cpu-unit-test pipeline (Job python3.6) passed all ut and it used a lower version of pip 21.2.4.

dong0321:
image

Seems to be the same issue. I am building release/0.10 with rocm6.0 dockerfile.

dong0321:
@RyoYang could please check if this need to be fixed also in release/0.10

dong0321:
Build successed after change to 2.3.0.

PR #636, opened 2024-08-08
Bug - Fix issues caused by omegaconf version update 

RyoYang:
**Description**
* Fix executor for Benchmark Execution Without Explicit Framework Field
* Fix mca misspell issue
* Timeout should be None if it is not explicated

### DIFF
--- a/superbench/runner/runner.py
+++ b/superbench/runner/runner.py
@@ -448,7 +448,7 @@ class SuperBenchRunner():
             mode.env.update({'SB_MODE_SERIAL_INDEX': mode.serial_index, 'SB_MODE_PARALLEL_INDEX': mode.parallel_index})
         logger.info('Runner is going to run %s in %s mode, proc rank %d.', benchmark_name, mode.name, mode.proc_rank)
 
-        timeout = self._sb_benchmarks[benchmark_name].get('timeout', 60)
+        timeout = self._sb_benchmarks[benchmark_name].get('timeout', None)
         if isinstance(timeout, int):
             timeout = max(timeout, 60)
 
--- a/superbench/runner/runner.py
+++ b/superbench/runner/runner.py
@@ -84,7 +84,7 @@ class SuperBenchRunner():
                     if 'proc_num' not in mode:
                         self._sb_benchmarks[name].modes[idx].proc_num = 8
                 elif mode.name == 'mpi':
-                    if 'machinefile' not in mode:
+                    if 'mca' not in mode:
                         self._sb_benchmarks[name].modes[idx].mca = {
                             'pml': 'ob1',
                             'btl': '^openib',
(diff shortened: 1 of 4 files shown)


superbench/runner/runner.py
```diff
@@ -84,7 +84,7 @@ def __validate_sb_config(self):    # noqa: C901
                     if 'proc_num' not in mode:
                         self._sb_benchmarks[name].modes[idx].proc_num = 8
                 elif mode.name == 'mpi':
-                    if 'machinefile' not in mode:
+                    if 'mca' not in mode:
```
yukirora, line 87: @RyoYang , could you please also add a test case for this?

superbench/runner/runner.py
```diff
@@ -448,7 +448,7 @@ def _run_proc(self, benchmark_name, mode, vars):
             mode.env.update({'SB_MODE_SERIAL_INDEX': mode.serial_index, 'SB_MODE_PARALLEL_INDEX': mode.parallel_index})
         logger.info('Runner is going to run %s in %s mode, proc rank %d.', benchmark_name, mode.name, mode.proc_rank)
 
-        timeout = self._sb_benchmarks[benchmark_name].get('timeout', 60)
+        timeout = self._sb_benchmarks[benchmark_name].get('timeout', None)
```
yukirora, line 451: @RyoYang , could you please also add a test case for this?

superbench/runner/runner.py
yukirora: and these 2 bugs in omegaconf upgrade PR seems no relationship with the title/description? please change the title or split the PR

RyoYang: Updated.

superbench/runner/runner.py
RyoYang: Updated.

yukirora (approved):

guoshzhao:
Could you please check if:
1. All configs used in Prod env have `timeout` set? https://msazure.visualstudio.com/One/_git/SuperBench?path=/config
2. Are these two bugs related with omegaconf version update?
If the answers are Yes, I am ok with this PR. 

If the answer is No for 1, then we changed the behavior of Prod.
If the answer is No for 2, maybe we need to split to two PRs?

Thanks Yang for the efforts.

RyoYang:
> Could you please check if:
> 
> 1. All configs used in Prod env have `timeout` set? https://msazure.visualstudio.com/One/_git/SuperBench?path=/config
> 2. Are these two bugs related with omegaconf version update?
>    If the answers are Yes, I am ok with this PR.
> 
> If the answer is No for 1, then we changed the behavior of Prod. If the answer is No for 2, maybe we need to split to two PRs?
> 
> Thanks Yang for the efforts.


1. The change addresses a bug in the previous PR where all benchmarks without a specified timeout were forcefully set to a 60-second timeout. For the production configuration, if the timeout is not set, it should default to None, meaning the benchmark should have no timeout.
2. Yes, these changes are related to the update of the omegaconf version.
Thanks a lot for the review.

abuccts (approved):

PR #664, opened 2024-11-14
Benchmarks: micro benchmarks - add nvbandwidth build

polarG:
**Description**
Add nvbandwidth build to repo


third_party/Makefile
```diff
@@ -239,3 +239,10 @@ ifneq (,$(wildcard msccl/tests/msccl-tests-nccl/Makefile))
 	mkdir -p $(SB_MICRO_PATH)/bin/msccl-tests-nccl && \
 	cp -r -v ./msccl/tests/msccl-tests-nccl/build/* $(SB_MICRO_PATH)/bin/msccl-tests-nccl/
 endif
+
+# Build nvbandwidth.
+nvbandwidth: sb_micro_path
+ifneq (,$(wildcard nvbandwidth/debian_install.sh))
```
abuccts, line 245: can you check the cmake version here?
```makefile
ifeq ($(shell echo $(cmake --version | grep -oP '\d+\.\d+')">=3.20" | bc -l), 1)
```

third_party/Makefile
```diff
@@ -239,3 +239,10 @@ ifneq (,$(wildcard msccl/tests/msccl-tests-nccl/Makefile))
...
+	cd ./nvbandwidth && ./debian_install.sh && cd ..
```
abuccts, line 246: maybe run `cmake . && make` directly? dependencies have already been installed in dockerfile and may be broken when re-install

polarG:
Need to clean up the code and will create another pr.

PR #665, opened 2024-11-16
Benchmarks: micro benchmarks - add nvbandwidth build

polarG:
**Description**
Add nvbandwidth build to repo


abuccts (approved):

third_party/Makefile
```diff
@@ -239,3 +239,16 @@ ifneq (,$(wildcard msccl/tests/msccl-tests-nccl/Makefile))
 	mkdir -p $(SB_MICRO_PATH)/bin/msccl-tests-nccl && \
 	cp -r -v ./msccl/tests/msccl-tests-nccl/build/* $(SB_MICRO_PATH)/bin/msccl-tests-nccl/
 endif
+
+# Build nvbandwidth.
+REQUIRED_CMAKE_VERSION := 3.20.0
+CMAKE_VERSION := $(shell cmake --version 2>/dev/null | head -n 1 | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
+nvbandwidth: sb_micro_path
+	@if [ "$(CMAKE_VERSION)" = "" ]; then \
+        echo "Skip nvbandwidth for CMake is not installed."; \
```
yzygitzh, line 248: I think we should expose thrid-party dependency issues in this file instead of skipping it. Maybe we don't need to check cmake version here.

yzygitzh, line 248: I think we should expose thrid-party dependency issues in this file instead of skipping it. Maybe we don't need to check cmake version here.

yzygitzh (changes requested):

third_party/Makefile
yzygitzh: Discussed with @abuccts offline. Seems nvbandwidth will only be enabled for h100 and above nv gpus. In that case, can we change this to only enabled when this target is explicitly enabled when calling make? https://github.com/microsoft/superbenchmark/blob/main/third_party/Makefile#L23 and https://github.com/microsoft/superbenchmark/blob/main/dockerfile/cuda12.2.dockerfile#L142 can be a good reference.

yzygitzh (approved):

PR #672, opened 2024-11-26
fix cache tests issues with cuda 12.4 and pytorch 2

dpower4:
Fix cache tests issues with cuda 12.4 and pytorch 2

https://pytorch.org/docs/stable/notes/cuda.html#cuda-memory-management:
```
PyTorch uses a caching memory allocator to speed up memory allocations. This allows fast memory deallocation without
device synchronizations. However, the unused memory managed by the allocator will still show as if used in nvidia-smi. 
You can use memory_allocated() and max_memory_allocated() to monitor memory occupied by tensors, and use 
memory_reserved() and max_memory_reserved() to monitor the total amount of memory managed by the caching allocator. 

Calling empty_cache() releases all unused cached memory from PyTorch so that those can be used by other GPU 
applications. However, the occupied GPU memory by tensors will not be freed so it can not increase the amount of GPU 
memory available for PyTorch.
```

### DIFF
--- a/tests/benchmarks/model_benchmarks/test_pytorch_base.py
+++ b/tests/benchmarks/model_benchmarks/test_pytorch_base.py
@@ -250,16 +250,36 @@ def test_pytorch_empty_cache():
     # Register mnist benchmark.
     BenchmarkRegistry.register_benchmark('pytorch-mnist', PytorchMNIST)
 
+    # Get initial memory reserved
+    init_res_memory = torch.cuda.memory_reserved()
+    assert (init_res_memory >= 0)
+
     # Test cache empty by manually calling torch.cuda.empty_cache().
     parameters = '--batch_size 32 --num_warmup 8 --num_steps 64 --model_action train'
     benchmark = PytorchMNIST('pytorch-mnist', parameters=parameters)
+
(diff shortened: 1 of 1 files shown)


tests/benchmarks/model_benchmarks/test_pytorch_base.py
```diff
@@ -250,16 +250,36 @@ def test_pytorch_empty_cache():
     # Register mnist benchmark.
     BenchmarkRegistry.register_benchmark('pytorch-mnist', PytorchMNIST)
 
+    # Get initial memory reserved
+    init_res_memory = torch.cuda.memory_reserved()
+    assert (init_res_memory >= 0)
```
polarG, line 255: Can memory_reserved() return a negative value? or this assert is unnecessary.

dpower4: left the check as the return was an int type.
But based on the [impl](https://github.com/pytorch/pytorch/blob/main/torch/cuda/memory.py#L204-L209), it is guaranteed to be a non negative integer.
removing the redundant check.

polarG (approved):
LGTM

PR #685, opened 2025-02-06
Bugfix: Avoid Unintended nvbandwidth Function Calls in All Benchmarks

polarG:
Root Cause:

1. '_get_all_test_cases()' was called in '_parser' while '_parser' was defined in the base class.
2.  in '_get_all_test_cases()', cmd path was not included.

Fix:

1. Remove '_get_all_test_cases()' from '_parser'.
2. Construct path for cmd.


guoshzhao (approved):

dpower4 (approved):
Lgtm

PR #687, opened 2025-02-18
Init latest python support.

polarG:
Added support for Python 3.11, 3.12 and 3.13.

yapf is not compatiable with python3.12+, so we disable yapf in py3.12 for now.
https://github.com/google/yapf/issues/1258
https://github.com/google/yapf/issues/1266


setup.py
```diff
@@ -51,8 +59,11 @@ def finalize_options(self):
 
     def run(self):
         """Fromat the code using yapf."""
```
dpower4, line 61: typo in Format.

setup.py
```diff
@@ -51,8 +59,11 @@ def finalize_options(self):
...
-        errno = os.system('python3 -m yapf --in-place --recursive --exclude .git --exclude .eggs .')
-        sys.exit(0 if errno == 0 else 1)
+        if sys.version_info[:2] == (3, 12):
```
dpower4, line 62: Adda TODO: here to enable it once yapf implements support for 3.12

dpower4 (approved):
LGTM, typo fix needed.

.azure-pipelines/cpu-unit-test.yml
polarG, line 1: @abuccts @guoshzhao Do we need to keep 3.7 and 3.8? or is it necessary to verify all the python version.? Can we only support the latest version in the new release. When user want to use old version, they can use the previous release.

dpower4: My two cents:  it will get complex to maintain backsupport in future releases. Since 3.8 security support also ended in Oct 2024, its better to remove 3.7 and 3.8 from future releases and update docs accordingly. As you suggested if the need is 3.8 or below, the users can rely on prev releases.

guoshzhao (approved):

setup.py
```diff
@@ -12,19 +12,21 @@
...
-    print(
-        '\033[93mTry update pip/setuptools versions, for example, '
-        'python3 -m pip install --upgrade pip wheel setuptools==65.7\033[0m'
-    )
-    raise
+print(f'Python {sys.version_info.major}.{sys.version_info.minor} detected.')
+if sys.version_info[:2] < (3, 11):
+    import pkg_resources
+    try:
+        pkg_resources.require(['pip>=18', 'setuptools>=45, <66'])
```
abuccts, line 23: is pip<18 compatible with python3.11+?

setup.py
```diff
@@ -131,10 +141,11 @@ def run(self):
         'Operating System :: POSIX',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3 :: Only',
-        'Programming Language :: Python :: 3.7',
-        'Programming Language :: Python :: 3.8',
```
abuccts, line 135: why remove 3.7 and 3.8?

setup.py
```diff
@@ -151,8 +162,11 @@ def run(self):
         'setuptools_scm',
     ],
     install_requires=[
-        'ansible_base>=2.10.9;os_name=="posix"',
-        'ansible_runner>=2.0.0rc1, <2.3.2',
+        'ansible;python_version>"3.10"',
+        'ansible_base>=2.10.9;python_version<="3.10"',
```
abuccts, line 166: why remove `os_name=="posix"`? can it be installed on windows?

setup.py
```diff
@@ -194,7 +208,7 @@ def run(self):
             'test': [
                 'flake8-docstrings>=1.5.0',
                 'flake8-quotes>=3.2.0',
-                'flake8>=3.8.4, <6.0.0',
+                'flake8>=3.8.4',
```
abuccts, line 211: is this compatible with python 3.7?

abuccts (changes requested):

.azure-pipelines/cpu-unit-test.yml
```diff
@@ -7,16 +7,15 @@ trigger:
...
-    #python-latest:
-    #  imageTag: '3'
-  maxParallel: 3
+    python-3.11:
+      imageTag: '3.11'
+    python-3.12:
+      imageTag: '3.12'
+    python-3.13:
+      imageTag: '3.13'
+  maxParallel: 4
```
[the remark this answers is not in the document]
polarG: There is no requirement to support 3.13. I did a test build based on 3.13 and nothing wrong with it. Then added it.

setup.py
polarG: From 3.11, the default pip version is > 22.

setup.py
polarG: 3.7 and 3.8 are end of life. So, there is no need to support them. If the user wants to use old version, they can choose the previous released SB.

.azure-pipelines/cpu-unit-test.yml
```diff
@@ -7,16 +7,15 @@ trigger:
...
-    # TODO
-    #python-latest:
-    #  imageTag: '3'
-  maxParallel: 3
+    python-3.11:
+      imageTag: '3.11'
+    python-3.12:
+      imageTag: '3.12'
+    python-3.13:
+      imageTag: '3.13'
```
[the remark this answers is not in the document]
polarG: Pls give more details of 'break old versions'. thanks.

setup.py
polarG: In our user doc, it has specified that SB can only run on Linux. So, I do not think it is necessary.

setup.py
polarG: Yes, flake8 >= 6.0.0 needs python >=3.8.1

.azure-pipelines/cpu-unit-test.yml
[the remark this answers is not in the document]
polarG: Used 3.7, 3.10 and 3.12.

setup.py
polarG: Added 3.7 back.

PR #689, opened 2025-02-26
Refactor: Enhance logging information for diagnosis rule op baseline errors.

jorgeesg:
Improves logging info for diagnosis rule op baseline errors. This allows developers to easily detect errors in their rule files as well as baseline files, improving end-user experience.

### DIFF
--- a/superbench/analyzer/diagnosis_rule_op.py
+++ b/superbench/analyzer/diagnosis_rule_op.py
@@ -132,7 +132,7 @@ class RuleOp:
                 baseline = rule['metrics'][metric]
                 if baseline is None or baseline == 0:
                     logger.log_and_raise(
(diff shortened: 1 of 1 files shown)


abuccts (approved):
Is it possible to add a test case in unit tests for this change? maybe use [unittest.TestCase.assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises)

jorgeesg:
> Is it possible to add a test case in unit tests for this change? maybe use [unittest.TestCase.assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises)

Added

polarG (approved):
LGTM

jorgeesg:
@microsoft-github-policy-service agree company="Microsoft"

yukirora (approved):

PR #692, opened 2025-03-07
Lint: Enhancement of ignoring errors for import pkg_resources

polarG:
This enhancement addresses an issue in mypy where it may report missing pkg_resources even when ignore_missing_imports = True is set and the package is installed. Adding this configuration ensures that pkg_resources is properly skipped during type checking.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -3,3 +3,7 @@ exclude = build/
 ignore_missing_imports = True
 scripts_are_modules = True
 warn_return_any = True
+
+[mypy-pkg_resources]
+ignore_missing_imports = True
+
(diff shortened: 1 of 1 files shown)


dpower4 (approved):

.mypy.ini
```diff
@@ -3,3 +3,7 @@ exclude = build/
 ignore_missing_imports = True
 scripts_are_modules = True
 warn_return_any = True
+
+[mypy-pkg_resources]
+ignore_missing_imports = True
+
```
abuccts, line 9: remove the trailing empty line

abuccts (approved):

PR #693, opened 2025-03-08
CI/CD - Update label in the ROCm image build

polarG:
Due to the matrix strategy’s default "fail-fast" setting. In GitHub Actions, when running a job with a matrix, the individual configurations run in parallel. By default, if one matrix job (for example, the one labeled "rocm6_2_rocm6_2_x_superbe") fails, the remaining parallel jobs are canceled automatically.

In our current build image pipeline, the arm64 build job always are canceled by the rocm build job. So, using a non-existent label in the job config to prevent rocm build job from scheduling for a temporary solution.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -53,7 +53,7 @@ jobs:
           dockerfile: rocm6.2.x
           tags: superbench/main:rocm6.2
           platforms: linux/amd64
-          runner: [self-hosted, rocm]
+          runner: [self-hosted, linux/amd64, rocm]
           build_args: "NUM_MAKE_JOBS=16"
     steps:
       - name: Checkout
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -23,7 +23,6 @@ jobs:
       contents: read
       packages: write
     strategy:
-      fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-image.yml
```diff
@@ -49,12 +49,6 @@ jobs:
           platforms: linux/amd64
           runner: ubuntu-latest
           build_args: "NUM_MAKE_JOBS=8"
-        - name: rocm6.2
-          dockerfile: rocm6.2.x
-          tags: superbench/main:rocm6.2
-          platforms: linux/amd64
-          runner: [self-hosted, linux/amd64]
```
abuccts, line 56: can you just add a non-exist label like rocm?

dpower4 (approved):

.github/workflows/build-image.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           dockerfile: rocm6.2.x
           tags: superbench/main:rocm6.2
           platforms: linux/amd64
-          runner: [self-hosted, linux/amd64]
+          runner: [self-hosted, rocm]
```
abuccts, line 56: can you keep the origin tags and just add a new one?

abuccts (commented):
pls update the title/description accordingly as well

abuccts (approved):

PR #703, opened 2025-04-23
Introduce --memPoolSize to trtexec

Akado2009:
**Description**
trtexec has a new parameter --memPoolSize, which was introduced as a substitute for --workspace (trtexec with workspace says that the parameters is DEPRECATED).
Link to official nvidia docs: [trtexec](https://docs.nvidia.com/deeplearning/tensorrt/latest/reference/command-line-programs.html)

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/tensorrt_inference_performance.py
+++ b/superbench/benchmarks/micro_benchmarks/tensorrt_inference_performance.py
@@ -104,7 +104,7 @@ class TensorRTInferenceBenchmark(MicroBenchmarkWithInvoke):
                 # build options
                 '--explicitBatch',
                 f'--optShapes=input:{input_shape}',
-                '--workspace=8192',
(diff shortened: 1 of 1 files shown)


superbench/benchmarks/micro_benchmarks/tensorrt_inference_performance.py
```diff
@@ -104,7 +104,7 @@ def _preprocess(self):
                 # build options
                 '--explicitBatch',
                 f'--optShapes=input:{input_shape}',
-                '--workspace=8192',
+                '–-memPoolSize=8G',
```
polarG, line 107: The --memPoolSize parameter in trtexec was introduced with TensorRT 8.5, which requires CUDA 11.4 or later.
Since we still support CUDA 11.1, we need to ensure compatibility with environments that do not have --memPoolSize available.

PR #705, opened 2025-05-02
Feat/py2503 build

dpower4:
- Updated to use the latest pytorch ngc container 25.03 as it has latest cublas 12.8.4 . 
- regressions with cublas spotted from 25.02 version: 12.8.3
- added cutlas arch as they ignore the 
- use recent cutlass release v3.9.1
- Condense cutlass build to 90;90a;100;100a from 12.8 docker images to reduce build time.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -224,7 +224,7 @@ setup(
                 'yapf==0.31.0',
             ],
             'torch': [
-                ''safetensors<=0.5.3',
+                'safetensors<=0.5.3',
                 'tokenizers<=0.20.3',
                 'torch>=1.7.0a0',
                 'torchvision>=0.8.0a0',
--- a/setup.py
+++ b/setup.py
@@ -183,7 +183,7 @@ setup(
         'openpyxl>=3.0.7',
         'packaging>=21.0',
         'pandas>=1.1.5',
-        'protobuf<=3.20.3',
+        'protobuf<=4.24.4',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -224,7 +224,7 @@ def run(self):
                 'yapf==0.31.0',
             ],
             'torch': [
-                'safetensors==0.4.5',
+                'safetensors<=0.5.3',
```
polarG, line 227: Since this is for the specific cuda dockerfile, could we use fix version for this?
'torch' has performance impact on the outputs.

third_party/Makefile
```diff
@@ -38,12 +38,12 @@ sb_micro_path:
 	mkdir -p $(SB_MICRO_PATH)/lib
 
 # Build cutlass.
-# for cuda 12.8 and later Build from commit 389e493 (3.8 release commit) for blackwell support
+# for cuda 12.8 and later Build from commit v3.9 (3.9 release commit) for blackwell support
 cuda_cutlass:
 ifeq ($(shell echo $(CUDA_VER)">=12.8" | bc -l), 1)
-	$(eval ARCHS := "75;80;86;89;90a;100;100a")
```
guoshzhao, line 44: We need to keep `80` for A100 support, since there are teams validating A100 with cuda 12.8 & 12.9.

dpower4:
closing this as we will move to 12.9 based build #716

PR #711, opened 2025-05-19
Benchmarks - Add FP4 GEMM FLOPS support for cublaslt_gemm benchmark

guoshzhao:
**Description**
Add FP4 precision support for cublaslt_gemm benchmark.

**Major Revision**
- Add new type `fp4e2m1` and `__nv_fp4_e2m1`.
- For FP4 matmul, precision of MatrixC (add) should be FP16, precision of MatricD (output) should be FP4, otherwise, it will not work.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/cublaslt_gemm/cublaslt_gemm.cu
+++ b/superbench/benchmarks/micro_benchmarks/cublaslt_gemm/cublaslt_gemm.cu
@@ -7,6 +7,7 @@
 
 #include <cuda_fp16.h>
 #include <cuda_fp8.h>
+#include <cuda_fp4.h>
 
 #include "cublaslt_utils.h"
 
@@ -16,6 +17,7 @@ using fp16 = half;
 using bf16 = nv_bfloat16;
 using fp8e4m3 = __nv_fp8_e4m3;
 using fp8e5m2 = __nv_fp8_e5m2;
+using fp4e2m1 = __nv_fp4_e2m1;
 using int8 = int8_t;
 
 struct Args {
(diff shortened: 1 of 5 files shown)


.azure-pipelines/cuda-unit-test.yml
```diff
@@ -11,7 +11,7 @@ pool:
   vmImage: ubuntu-latest
 
 container:
-  image: nvcr.io/nvidia/pytorch:24.03-py3
+  image: nvcr.io/nvidia/pytorch:25.02-py3
```
polarG, line 14: We should also remove the line 27, since the default python3 in this image is 3.12. And SB do not need special version of setuptools for the last python.

polarG: Apply control in Makefile

superbench/benchmarks/micro_benchmarks/cublaslt_gemm/cublaslt_gemm.cu
```diff
@@ -102,32 +108,39 @@ template <typename T> cudaDataType_t get_datatype() {
...
                        int iter_autotune, int warmup_autotune) {
     // init matrix
     Ta *matrix_a = nullptr;
     Tb *matrix_b = nullptr;
+    Tc *matrix_c = nullptr;
     Tout *matrix_out = nullptr;
     batch = std::max<size_t>(batch, 1);
     cudaMalloc(&matrix_a, m * k * batch * sizeof(Ta));
     cudaMalloc(&matrix_b, k * n * batch * sizeof(Tb));
+    cudaMalloc(&matrix_c, m * n * batch * sizeof(Tc));
```
polarG, line 131: where are they released?

guoshzhao: It is at the end of this function, line 178~181.

PR #712, opened 2025-05-21
microbenchmark - CPU Stream Benchmark Revise

polarG:
In the current implementation, the CPU‑stream benchmark code renames the binary before the microbench base class can verify its existence, causing the default‐binary check to fail. 

This PR adds a “default” binary—built with the standard compile parameters—so that the base class can always find and validate it. Once the default binary is in place, the CPU‑stream code will rename it as needed and re‑check its presence before running the benchmark.

The PR also enable CPU stream in the default settings.


third_party/Makefile
```diff
@@ -23,12 +23,12 @@ all: cuda rocm
...
 rocm: common rocm_perftest rocm_rccl_tests rocm_rocblas rocm_bandwidthTest rocm_hipblaslt megatron_deepspeed apex_rocm
-cpu: common cpu_perftest cpu_stream
-common: fio
+cpu: common cpu_perftest
+common: fio cpu_stream
 
 # non aarch64 specific targets
 ifneq ($(shell uname -m), aarch64)
-common: fio cpu_hpl
+common: cpu_hpl
```
abuccts, line 31: why remove fio?

abuccts (commented):
pls fix the build

guoshzhao (approved):

third_party/Makefile
polarG: Common already have assigned value fio and there no need to append again. So removed it.

PR #717, opened 2025-06-03
Benchmarks - Remove arch 70(V100) and 75(T4) for cutlass building

guoshzhao:
**Description**
Remove arch 70(V100) and 75(T4) for cutlass building


dpower4 (changes requested):
it might be better to keep the support for 75 for cuda 11.8 and below.
for the 12.8 onwards, we can  start support from 90 and beyond

guoshzhao:
> it might be better to keep the support for 75 for cuda 11.8 and below. for the 12.8 onwards, we can start support from 90 and beyond

I noticed you already have a PR removing arch for cutlass building. I will close this one. Thanks!

PR #722, opened 2025-07-10
Fix deepseek benchmark names conflict.

polarG:
Summary
Deepseek currently registers two benchmarks using the same identifier, causing a naming conflict.

Change
Rename each benchmark so that both have unique, descriptive names.


abuccts (commented):
`release/*` branch is protected, please rename your develop branch

polarG:
Close this PR to correct the source branch.

PR #724, opened 2025-07-21
Bugfix: deepseek benchmark names conflict.

polarG:
Summary
Deepseek currently registers two benchmarks using the same identifier, causing a naming conflict.

Change
Rename each benchmark so that both have unique, descriptive names.

### DIFF
--- a/superbench/benchmarks/model_benchmarks/megatron_gpt3.py
+++ b/superbench/benchmarks/model_benchmarks/megatron_gpt3.py
@@ -749,7 +749,7 @@ BenchmarkRegistry.register_benchmark(
     'megatron-deepseek-v2', MegatronGPT, parameters='--model=deepseek', platform=Platform.ROCM
 )
 BenchmarkRegistry.register_benchmark(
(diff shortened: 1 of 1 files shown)


guoshzhao (approved):

superbench/benchmarks/model_benchmarks/megatron_gpt3.py
```diff
@@ -749,7 +749,7 @@ def _cal_params_count(self):
     'megatron-deepseek-v2', MegatronGPT, parameters='--model=deepseek', platform=Platform.ROCM
 )
 BenchmarkRegistry.register_benchmark(
-    'megatron-deepseek-v2',
+    'megatron-deepseek-v2-lite',
```
yukirora, line 752: where is the another deepseek benchmark?

polarG: Pls refer line 748-750. It is still using the original name 'megatron-deepseek-v2'.

PR #731, opened 2025-08-28
Aishwaryatonpe/deterministic training

Aishwarya-Tonpe:
Support for deterministic training and reproducible logging to all PyTorch model benchmarks in SuperBench (BERT, GPT2, LLaMA, LSTM, CNN, Mixtral).

Deterministic mode: Makes sure model runs are consistent every time by fixing random seeds, turning off TF32, and using stable math operations.
Log generation: Saves key info like loss and activation stats during training.
Log comparison: Lets you compare a new run with a previous one to check if they match.
New command-line options:

--deterministic
--generate-log
--compare-log
--log-path
--check-frequency

Changes - 

Updated pytorch_base.py to handle deterministic settings, logging, and comparisons.
Added a new example script: pytorch_deterministic_example.py
Added a test file: test_pytorch_determinism_all.py to verify everything works as expected.

Usage - 

Run with --deterministic --generate-log to create a reference log.
Run again with --compare-log to check if the new run matches the reference.
Make sure CUBLAS_WORKSPACE_CONFIG is set and all parameters stay the same between runs.


examples/benchmarks/pytorch_deterministic_example.py
```diff
@@ -0,0 +1,98 @@
...
+    parser.add_argument(
+        '--deterministic-seed',
+        type=int,
+        default=42,
+        help='Seed for deterministic training.',
+    )
+    args = parser.parse_args()
+
+    parameters = DEFAULT_PARAMS[args.model]
+    parameters = parameters.replace('--deterministic_seed', '--deterministic_seed')
```
guoshzhao: why need this line?

superbench/benchmarks/base.py
```diff
@@ -119,6 +122,55 @@ def parse_args(self, ignore_invalid=False):
...
+                else:
+                    converted.append(Precision(v))
+            return converted
+        else:
+            if isinstance(value, Precision):
+                return [value]
+            else:
+                return [Precision(value)]
+
+    def _check_unknown_args(self, unknown):
```
guoshzhao, line 117: Where is this function used?

superbench/benchmarks/base.py
```diff
@@ -119,6 +122,55 @@ def parse_args(self, ignore_invalid=False):
 
         return ret, args, unknown
 
+    def _parse_args_override_step(self, args):
```
guoshzhao: Please add docstring for this function.

superbench/benchmarks/base.py
```diff
@@ -119,6 +122,55 @@ def parse_args(self, ignore_invalid=False):
...
+                        if key == 'precision' and Precision is not None:
+                            setattr(args, key, self._convert_precision_value(value, Precision))
+                        else:
+                            setattr(args, key, value)
+                logger.info(f'Arguments overridden from compare_log metadata for determinism. New Arguments: {args}')
+            except Exception as e:
+                logger.info(f'Failed to override args from compare_log metadata: {e}')
+        return args
+
+    def _convert_precision_value(self, value, Precision):
```
guoshzhao: Please add docstring for all functions.

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,243 @@ def __init__(self, name, parameters=''):
...
+            random.seed(self._args.deterministic_seed)
+            if torch.cuda.is_available():
+                torch.cuda.manual_seed(self._args.deterministic_seed)
+                torch.cuda.manual_seed_all(self._args.deterministic_seed)
+        torch.use_deterministic_algorithms(True, warn_only=False)
+        torch.backends.cudnn.deterministic = True
+        torch.backends.cudnn.benchmark = False
+        # Disable TF32 to remove potential numerical variability
+        try:
+            torch.backends.cuda.matmul.allow_tf32 = False
```
guoshzhao, line 73: Add the same log as cudnn? For example, `logger.info('Failed to disable TF32 in cuda matmul')`

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,243 @@ def __init__(self, name, parameters=''):
...
+        try:
+            v = float(loss.detach().item()) if hasattr(loss, 'detach') else float(loss)
+        except Exception:
+            v = None
+        # Periodic fingerprint logging
+        if getattr(self._args, 'deterministic', False) and (curr_step % check_frequency == 0):
+            # 1) Loss fingerprint (only at fingerprinting frequency)
+            try:
+                if 'loss' in periodic and v is not None:
+                    periodic['loss'].append(v)
```
guoshzhao: Is it fine that the length of `periodic['loss']` may be different (shorter) with the length of `periodic['step']`?

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,243 @@ def __init__(self, name, parameters=''):
...
+                logger.info(f'Loss at step {curr_step}: {v}')
+                periodic['step'].append(curr_step)
+            except Exception:
+                pass
+            # 2) Tiny activation fingerprint: mean over logits for sample 0
+            try:
+                if logits is not None:
+                    act_mean = (
```
guoshzhao: Same questions here `periodic['loss']`, `periodic['step']` and `periodic['act_mean']` may have different length. Can we know the loss is from which step?

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,243 @@ def __init__(self, name, parameters=''):
...
+                        float(logits[0].detach().float().mean().item())
+                        if hasattr(logits[0], 'detach') else float(logits[0])
+                    )
+                    logger.info(f'ActMean at step {curr_step}: {act_mean}')
+                    periodic['act_mean'].append(act_mean)
+            except Exception:
+                pass
+
+    def _finalize_periodic_logging(self, duration, periodic, info_key='loss'):
```
guoshzhao: why put `duration` into this function as it has not been used?

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,243 @@ def __init__(self, name, parameters=''):
...
+            default=False,
+            help='Save fingerprint log to file.',
+        )
+        self._parser.add_argument(
+            '--log-path',
+            '--log_path',
+            dest='log_path',
+            type=str,
+            default=None,
+            help='Path to save or load fingerprint log.',
```
guoshzhao: Looks this argument is not used when `load` log in your example:
```
--model <model_from_MODEL_CHOICES> --generate-log --log-path ./outputs/determinism_ref.json
--model <model_from_MODEL_CHOICES> --compare-log ./outputs/determinism_ref.json
```

Should we unify the usage? Like
```
--model <model_from_MODEL_CHOICES> --compare-log --log-path ./outputs/determinism_ref.json
```

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -343,3 +586,29 @@ def _timer(self):
         if self._gpu_available:
             torch.cuda.synchronize()
         return time.time()
+
+    def _process_info(self, model_action, precision, info):
```
guoshzhao: Where is this function called?

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -94,6 +94,10 @@ def _generate_dataset(self):
         Return:
             True if dataset is created successfully.
         """
+        # Seed before dataset generation when deterministic
+        if getattr(self._args, 'deterministic', False) and hasattr(self._args, 'deterministic_seed'):
+            torch.manual_seed(self._args.deterministic_seed)
```
guoshzhao: Looks `torch.manual_seed(self._args.deterministic_seed)` is called twice in both `_generate_dataset` and `_create_model`'s `_enable_deterministic_training()`.
Can we just call `_enable_deterministic_training()` in `_preprocess` of `pytorch_base` class, then we don't need to call it in each pytorch model file.

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -109,6 +113,10 @@ def _create_model(self, precision):
         Args:
             precision (Precision): precision of model and input data, such as float32, float16.
         """
+        # Enable deterministic training if requested
+        if getattr(self._args, 'deterministic', False):
```
guoshzhao: Can we remove it once we call _enable_deterministic_training() in _preprocess of pytorch_base class?

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -180,17 +194,18 @@ def _train_step(self, precision):
...
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, logits, periodic, check_frequency)
                     self._log_step_time(curr_step, precision, duration)
                 if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                    return self._finalize_periodic_logging(duration, periodic)
```
guoshzhao: Don't need to pass `duration` into this function?

superbench/benchmarks/model_benchmarks/pytorch_base.py
Aishwarya-Tonpe: removed the `log-path` argument so that now there are only 2 arguments,` generate-log` and `compare-log`, we can optionally provide a file path to the `generate-log` argument and the logs will be saved at that location

superbench/benchmarks/model_benchmarks/pytorch_bert.py
Aishwarya-Tonpe: Tried the above approach, however it does not work and does not run a deterministic run possibly due to the following reasons - 
Moving the `torch.manual_seed(self._args.deterministic_seed`) into `pytorch_base _preprocess()` breaks reproducibility because RNG state must be set before any RNG-consuming work. Many models call generate_dataset() / create_model() before _preprocess() runs, so a central seed there is already too late.

superbench/benchmarks/model_benchmarks/pytorch_bert.py
Aishwarya-Tonpe: Passed this here as it already existed in the code, to keep the code persistent with the existing code and not make any changes unrelated to the SDC check functionality, I passed `duration`

superbench/benchmarks/model_benchmarks/pytorch_base.py
Aishwarya-Tonpe: handled this my appending `None `values for when loss might be skipped, same for act_mean

superbench/benchmarks/model_benchmarks/pytorch_base.py
Aishwarya-Tonpe: handled this my appending `None `values for when act_mean might be skipped

superbench/benchmarks/base.py
Aishwarya-Tonpe: Looks like it wasn't utilized even though it was created to fix the method too complex check, have utilized it now where it was meant to be.

superbench/benchmarks/model_benchmarks/pytorch_bert.py
Aishwarya-Tonpe: Passed it here for the following reasons - 
1. To keep it consistent with the existing code, as the original function returned `duration` before I added any changes related to SDC capability
2. It's possibly being used here - https://github.com/microsoft/superbenchmark/blob/main/superbench/benchmarks/model_benchmarks/model_base.py#L276
where it unpacks `step_times = self._train_step(precision)`
 ```
info = step_times[1]
step_times = step_times[0]
```

File: superbench/benchmarks/model_benchmarks/pytorch_base.py (hunk 8 above)
Comment (Aishwarya-Tonpe) [reply superbench/benchmarks/model_benchmarks/pytorch_base.py]: Passed it here for even though the function does not utilise it but so that it returns it for the following reasons -

To keep it consistent with the existing code, as the original function returned `duration `before I added any changes related to SDC capability
It's possibly being used here - https://github.com/microsoft/superbenchmark/blob/main/superbench/benchmarks/model_benchmarks/model_base.py#L276
where it unpacks `step_times = self._train_step(precision)`
```
info = step_times[1]
step_times = step_times[0]
```

File: superbench/benchmarks/model_benchmarks/pytorch_bert.py (hunk 13 above)
Comment (guoshzhao) [reply superbench/benchmarks/model_benchmarks/pytorch_bert.py]: Oh, what I mean is you can change to 
` return (duration, self._finalize_periodic_logging(periodic))`
Since `duration` is not needed inside `_finalize_periodic_logging()`

File: superbench/benchmarks/model_benchmarks/pytorch_base.py [hunk 14]
```diff
@@ -35,10 +44,256 @@ def __init__(self, name, parameters=''):
...
         self._gpu_available = not self._args.no_gpu and torch.cuda.is_available()
 
+    def _enable_deterministic_training(self):
+        """Enable deterministic training settings for reproducible results."""
+        if hasattr(self._args, 'deterministic_seed'):
+            torch.manual_seed(self._args.deterministic_seed)
+            random.seed(self._args.deterministic_seed)
+            if torch.cuda.is_available():
+                torch.cuda.manual_seed(self._args.deterministic_seed)
+                torch.cuda.manual_seed_all(self._args.deterministic_seed)
```
guoshzhao, line 67: Looks ` torch.cuda.manual_seed_all(self._args.deterministic_seed)` is enough here. Can we remove `torch.cuda.manual_seed(self._args.deterministic_seed)`?

superbench/benchmarks/model_benchmarks/pytorch_bert.py
guoshzhao: The current call sequence looks like:
> 
> 	    _preprocess()
> 		    add_parser_arguments()
> 		    parse_args()
> 		    _judge_gpu_availability()
> 		    _init_distributed_setting()
> 		    _generate_dataset()
> 		    _init_dataloader()
> 	
> 	    _benchmark()
> 		    __train() or __inference()
> 		    _create_model()
> 		    _create_optimizer()
> 		    _train_step()
> 		    _process_info()
> 		    __process_model_result()
> 		
> 	    _postprocess()

`_generate_dataset()` is the second to the last function call inside `_preprocess()`, and `create_model()` is called inside `_benchmark()` which is after `_preprocess()`. 

Can we define one empty abstract function such as `set_deterministic_seed()` and call it before `_generate_dataset()` inside `model_base.py`. Then in `pytorch_base.py`, we can implement this function.
In this way, we can make sure RNG state is set before `generate_dataset()`. Please let me know what do you think. Thanks.

superbench/benchmarks/base.py
```diff
@@ -110,14 +110,95 @@ def parse_args(self, ignore_invalid=False):
...
 
+        if args is not None and 'compare_log' in [a.dest for a in self._parser._actions]:
+            args = self._parse_args_override_step(args)
+
         ret = True
+        ret = self._check_unknown_args(unknown)
+
+        return ret, args, unknown
+
+    def _parse_args_override_step(self, args):
```
polarG: Why we need this new func since there is only inside call to another function.

superbench/benchmarks/base.py
```diff
@@ -110,14 +110,95 @@ def parse_args(self, ignore_invalid=False):
...
+                    Precision = None
+                for key, value in metadata.items():
+                    if hasattr(args, key):
+                        if key == 'precision' and Precision is not None:
+                            setattr(args, key, self._convert_precision_value(value, Precision))
+                        else:
+                            setattr(args, key, value)
+                logger.info(f'Arguments overridden from compare_log metadata for determinism. New Arguments: {args}')
+            except Exception as e:
+                logger.info(f'Failed to override args from compare_log metadata: {e}')
```
polarG: is logger.warning/error a better choice?

polarG: when exception occur, is code continue? or stopped here?

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,256 @@ def __init__(self, name, parameters=''):
...
+            curr_step (int): Current training step.
+            loss (torch.Tensor or float): Loss value for this step.
+            logits (torch.Tensor or float): Logits output for this step (sample 0).
+            periodic (dict): Dictionary to store periodic fingerprints ('loss', 'act_mean', 'step').
+            check_frequency (int): Frequency for fingerprint logging.
+        """
+        # Record per-step loss for determinism checks (for full history)
+        try:
+            v = float(loss.detach().item()) if hasattr(loss, 'detach') else float(loss)
+        except Exception:
```
polarG: Add logger to detail exception.

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,256 @@ def __init__(self, name, parameters=''):
...
+                # Ensure the lists exist and remain index-aligned by appending
+                # a placeholder (None) when a measurement is unavailable.
+                if 'loss' in periodic and isinstance(periodic['loss'], list):
+                    periodic['loss'].append(v if v is not None else None)
+                else:
+                    periodic['loss'] = [v if v is not None else None]
+
+                logger.info(f'Loss at step {curr_step}: {v}')
+                periodic.setdefault('step', []).append(curr_step)
+            except Exception:
```
polarG: Add logger to detail exception.

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,256 @@ def __init__(self, name, parameters=''):
...
+                        float(logits[0].detach().float().mean().item())
+                        if hasattr(logits[0], 'detach') else float(logits[0])
+                    )
+                    logger.info(f'ActMean at step {curr_step}: {act_mean}')
+                    periodic.setdefault('act_mean', []).append(act_mean)
+                else:
+                    # Keep lists aligned by appending None when activation not available
+                    periodic.setdefault('act_mean', []).append(None)
+            except Exception:
+                # On exception preserve alignment by ensuring keys exist
```
polarG: Add logger to detail exception.

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -182,17 +185,18 @@ def _train_step(self, precision):
...
                 loss.backward()
                 self._optimizer.step()
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, logits, periodic, check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
```
guoshzhao, line 194: Will the original behavior be changed if you put `if self._is_finished(curr_step, end, check_frequency):` inside the block of `if curr_step > self._args.num_warmup:`?
Previously, it can be early stopped during the warmup period.

Aishwarya-Tonpe: It does not entirely have a major impact but made this change to avoid stopping the benchmark while it's still in the warmup period so that measurements and logging stay consistent.
however, it does it prevents the benchmark from being stopped by `is_finished` during the warmup period.
I can move it back outside

superbench/benchmarks/model_benchmarks/pytorch_mixtral_impl.py
```diff
@@ -144,46 +164,55 @@ def _create_model(self, precision):
...
             self._target = self._target.cuda()
 
-        return True
+    def _assign_metadata_safe(self, precision):
+        try:
+            self._assign_model_run_metadata(
+                precision, extra_keys=['num_key_value_heads', 'max_position_embeddings', 'router_aux_loss_coef']
+            )
+        except Exception:
+            pass
```
polarG: Add logger here?

superbench/benchmarks/base.py
Aishwarya-Tonpe: The code does not stop, it continues. Yes, a warning logger would be better

polarG (approved):

docs/user-tutorial/benchmarks/model-benchmarks.md
```diff
@@ -34,6 +34,18 @@ For inference, supported percentiles include
 
 **New: Support fp8_hybrid and fp8_e4m3 precision for BERT models.**
 
+**New: SDC Support**
+SuperBench now supports SDC to ensure reproducibility across runs. This includes fixed seeds and deterministic algorithms. To enable SDC, the following flags and environment variables must be set:
+
+- **Flags:**
+  - `--deterministic`: Enables deterministic computation.
```
abuccts: please use verb or noun for argument, e.g., enable-determinism

docs/user-tutorial/benchmarks/model-benchmarks.md
```diff
@@ -34,6 +34,18 @@ For inference, supported percentiles include
...
+  - `--deterministic_seed <seed>`: Sets the seed for reproducibility.
```
abuccts: random-seed?

docs/user-tutorial/benchmarks/model-benchmarks.md
```diff
@@ -34,6 +34,18 @@ For inference, supported percentiles include
...
+  - `--generate_log` : Generates the log file that can be used as reference for comparison
```
abuccts: 1. what's current behavior of --generate-log in distributed training? seems there will be race conditions for multiple ranks writing the same log file
2. I think you only record loss/activation every n steps, could you generate these into the metrics file similar as current performance results, rather than using a new log file?

docs/user-tutorial/benchmarks/model-benchmarks.md
```diff
@@ -34,6 +34,18 @@ For inference, supported percentiles include
...
+  - `--compare_log <path>`: Specifies the path to the reference log for comparison.
```
abuccts: if the comparison necessary? if the loss etc. are one of the metrics, it can be separately compared like current throughput etc.

docs/user-tutorial/benchmarks/model-benchmarks.md
```diff
@@ -34,6 +34,18 @@ For inference, supported percentiles include
...
+
+- **Environment Variables:**
+  - `CUBLAS_WORKSPACE_CONFIG=:4096:8`: Ensures deterministic behavior in cuBLAS.
```
abuccts: I think you should set this in code when determinism feature is enabled rather than asking user to set it separately?

superbench/benchmarks/base.py
```diff
@@ -110,14 +110,84 @@ def parse_args(self, ignore_invalid=False):
                 logger.error('Invalid argument - benchmark: {}, message: {}.'.format(self._name, str(e)))
                 return False, None, []
 
+        if args is not None and 'compare_log' in [a.dest for a in self._parser._actions]:
+            args = self._override_args_with_compare_log(args)
```
abuccts: why determinism feature needs to override args?
for determinism, I think user just runs the model with same config for two or multiple times and compare whether the metrics (loss etc.) are same across runs

superbench/benchmarks/base.py
```diff
@@ -110,14 +110,84 @@ def parse_args(self, ignore_invalid=False):
...
+
         ret = True
+        ret = self._check_unknown_args(unknown)
```
abuccts, line 113: the first line in unnecessary

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -5,10 +5,12 @@
 
 import os
 from datetime import timedelta
+import random
 import time
 
 import torch
 import transformers
+
```
abuccts: remove

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -17,8 +19,15 @@
...
 from superbench.common.utils import logger
-from superbench.benchmarks import Framework, ReturnCode, DistributedBackend, DistributedImpl
+from superbench.benchmarks import (
+    Framework,
+    ReturnCode,
+    DistributedBackend,
+    DistributedImpl,
+)
 from superbench.benchmarks.model_benchmarks.model_base import Optimizer, ModelBenchmark
+from torch.backends.cuda import sdp_kernel
```
abuccts: pls put the third party dependencies, i.e., torch, into the second import block

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,261 @@ def __init__(self, name, parameters=''):
...
+        Args:
+            precision: Model precision (can be enum or string).
+            extra_keys: List of additional argument keys to include in metadata.
+            self._args: Benchmark arguments containing model configuration.
+
+        Returns:
+            None
+        """
+        # Common metadata keys
+        metadata = {
```
abuccts: I think these metadata values are decided by configuration, what if there exist conflicts?

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,261 @@ def __init__(self, name, parameters=''):
...
+            '--generate-log',
+            nargs='?',
+            const=True,
+            default=False,
+            type=str,
+            help='Save fingerprint log to file. Optionally specify a path to save the log.'
+        )
+        self._parser.add_argument(
+            '--compare-log',
+            '--compare_log',
```
abuccts: duplicate?

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,261 @@ def __init__(self, name, parameters=''):
...
+                log_path = gen_arg
+            if not log_path:
+                model = getattr(
+                    self._args,
+                    'model_name',
+                    self._name if hasattr(self, '_name') else 'model',
+                )
+                timestamp = time.strftime('%Y%m%d_%H%M%S')
+                os.makedirs('./outputs', exist_ok=True)
+                log_path = f'./outputs/model_run_{model}_{timestamp}.json'
```
abuccts: this is hard coded, please write necessary numbers in metrics rather than a new file

superbench/benchmarks/model_benchmarks/pytorch_cnn.py
```diff
@@ -108,17 +111,18 @@ def _train_step(self, precision):
...
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
                     # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, output, periodic, check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                    if self._is_finished(curr_step, end, check_frequency):
```
abuccts: this will change the behavior when the running will be stopped by duration rather than step number, please revert the indent

superbench/benchmarks/model_benchmarks/pytorch_cnn.py
```diff
@@ -132,6 +136,7 @@ def _inference_step(self, precision):
         """
         duration = []
         curr_step = 0
+        check_frequency = self._args.check_frequency
```
abuccts: why not use self._args.check_frequency directly if it's a constant number

tests/benchmarks/model_benchmarks/test_pytorch_determinism_all.py
```diff
@@ -0,0 +1,161 @@
...
+    # Log-file generation and contents
+    assert os.path.exists(log_path)
+    with open(log_path, 'r') as f:
+        data = json.load(f)
+    assert 'schema_version' in data
+    assert 'metadata' in data
+    assert 'per_step_fp32_loss' in data
+    assert 'fingerprints' in data
+    assert isinstance(data['per_step_fp32_loss'], list)
+    assert isinstance(data['fingerprints'], dict)
```
abuccts: is it possible to compare the detailed values here if it's deterministic?

tests/benchmarks/model_benchmarks/test_pytorch_determinism_all.py
```diff
@@ -0,0 +1,161 @@
...
+"""Unified test for deterministic fingerprinting across all major PyTorch model benchmarks."""
+
+from tests.helper import decorator
+import os
+import tempfile
+import json
+import pytest
+from superbench.benchmarks import BenchmarkRegistry, Platform, Framework, ReturnCode
+
+os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'
```
abuccts: pls set this in code rather than tests

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -35,10 +44,261 @@ def __init__(self, name, parameters=''):
...
+        preprocess_ok = super()._preprocess()
+        if not preprocess_ok:
+            return False
+        # Deterministic setup is handled centrally in set_deterministic_seed() which
+        # is invoked earlier in the model-base preprocess before dataset creation.
+        if getattr(self._args, 'deterministic', False):
+            self._handle_deterministic_log_options()
+        return True
+
+    def set_deterministic_seed(self):
```
abuccts, line 257: It seems most new functions here should be defined in model_base rather pytorch_base, you just have the pytorch related implementation

superbench/benchmarks/model_benchmarks/pytorch_base.py
Aishwarya-Tonpe: not a duplicate, have added functionality to take in compare-log (hyphen) as well as compare_log (underscore) since they are used interchangeably a lot of times to avoid unnecessary error, can remove support for one if needed

superbench/benchmarks/model_benchmarks/pytorch_base.py
Aishwarya-Tonpe: Currently, determinism support is only implemented for PyTorch models, so I placed all the determinism-related functions in pytorch_base.py.
However, if it's recommended to place them in model_base, I can move the functions to that file. Please let me know

superbench/benchmarks/base.py
Aishwarya-Tonpe: The purpose of override feature is to ensure that the user is using the same Metadata or the comparison would be void or incorrect.
Also, the comparison can be made between 2 different machines and not just multiple runs on the same machine, in that case there is a fair chance that the user accidentally uses a different configuration or Metadata. Overriding the args acts as a safety net and ensures that correct Metadata is being used for both the runs ensuring a valid comparison. 
If there's a mismatch in critical parameters, it's better to fail early with a clear error rather than showing "determinism failed" when the real issue is different configurations.

docs/user-tutorial/benchmarks/model-benchmarks.md
Aishwarya-Tonpe: The comparison serves a different purpose than performance metrics. The feature was designed this way to ensure exact equality and more importantly to generate a reference log (golden data) once and validate all subsequent runs across different machines and runs against it. 
I can integrate fingerprints into the standard metrics file (fixing the race condition), while keeping the exact-match comparison as a separate validation.

docs/user-tutorial/benchmarks/model-benchmarks.md
Aishwarya-Tonpe: Done

docs/user-tutorial/benchmarks/model-benchmarks.md
Aishwarya-Tonpe: Had initially named the flag as random-seed, but got a suggestion to rename it to deterministic_seed as that's more closely related to this feature.

docs/user-tutorial/benchmarks/model-benchmarks.md
Aishwarya-Tonpe: Implemented this approach

docs/user-tutorial/benchmarks/model-benchmarks.md
Aishwarya-Tonpe: Implemented and updated this

tests/benchmarks/model_benchmarks/test_pytorch_determinism_all.py
Aishwarya-Tonpe: since the test calls compare logic, it kind of compared the values 1-1. do you want assertions on 1-1 values compaisons in the test file itself?

docs/user-tutorial/benchmarks/model-benchmarks.md
```diff
@@ -34,6 +34,19 @@ For inference, supported percentiles include
...
 **New: Support fp8_hybrid and fp8_e4m3 precision for BERT models.**
 
+**New: Deterministic Training Support**
+SuperBench now supports deterministic training to ensure reproducibility across runs. This includes fixed seeds and deterministic algorithms. To enable deterministic training, the following flags and environment variables must be set:
+
+- **Flags:**
+  - `--enable-determinism`: Enables deterministic computation for reproducible results.
+  - `--deterministic_seed <seed>`: Sets the seed for reproducibility.
+  - `--generate_log` : Boolean flag that stores comparison metrics in the results file
+  - `--compare_log <results_file_path>`: Specifies the path to the reference file for comparison.
```
abuccts: unify them to use either underscore or dash?

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -30,15 +38,350 @@ def __init__(self, name, parameters=''):
...
+        if self._model_run_metadata:
+            if self._global_rank is not None:
+                metric_name = f'metadata_rank{self._global_rank}'
+            else:
+                metric_name = 'metadata'
+            # Use False for log_raw_data to save in result object, not log file
+            self._result.add_raw_data(metric_name, self._model_run_metadata, False)
+
+    def _save_consolidated_deterministic_results(self):
+        """Gather deterministic data from all ranks and save to results-summary (rank 0 only).
```
abuccts: all results from all ranks will be aggregated to control node by runner after benchmarks, I don't think this function is necessary

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -30,15 +38,350 @@ def __init__(self, name, parameters=''):
...
+            # Non-distributed: data already in result, nothing to consolidate
+            logger.info('Deterministic results stored in results')
+
+    def _compare_deterministic_results(self):
+        """Compare current deterministic metrics with reference results file.
+
+        Loads the reference results.json file and compares deterministic metrics
+        (loss, activation mean) per-rank to verify reproducibility.
+        """
+        import torch.distributed as dist
```
abuccts: why not import at the beginning

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -30,15 +38,350 @@ def __init__(self, name, parameters=''):
...
+                    f'Determinism check FAILED. Mismatched metrics:\n' + '\n'.join(mismatches)
+                )
+        except (FileNotFoundError, ValueError) as e:
+            has_failure = True
+            failure_msg = str(e)
+
+        # Synchronize failure status across all ranks in distributed mode
+        if self._args.distributed_impl == DistributedImpl.DDP:
+            # Convert failure status to tensor for all_reduce
+            import torch
```
abuccts: torch is also imported in this file, why import again?

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -30,15 +38,350 @@ def __init__(self, name, parameters=''):
...
+            failure_tensor = torch.tensor([1 if has_failure else 0], dtype=torch.int32, device='cuda')
+            dist.all_reduce(failure_tensor, op=dist.ReduceOp.MAX)
```
abuccts: will this work for cpu mode?

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -182,17 +184,18 @@ def _train_step(self, precision):
...
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, logits, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                    if self._is_finished(curr_step, end):
+                        return duration, self._finalize_periodic_logging(periodic)
```
abuccts: this will change the behavior when the running will be stopped by duration rather than step number

superbench/benchmarks/model_benchmarks/pytorch_cnn.py
```diff
@@ -108,17 +110,18 @@ def _train_step(self, precision):
...
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
                     # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, output, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                    if self._is_finished(curr_step, end):
+                        return duration, self._finalize_periodic_logging(periodic)
```
abuccts: same, this will change the behavior when the running will be stopped by duration rather than step number

superbench/benchmarks/model_benchmarks/pytorch_gpt2.py
```diff
@@ -158,11 +160,11 @@ def _train_step(self, precision):
             precision (Precision): precision of model and input data, such as float32, float16.
 
         Return:
-            The step-time list of every training step.
+           A tuple of (step_times_ms, info) of every training step.
```
abuccts: missing one space in indent

superbench/benchmarks/model_benchmarks/pytorch_gpt2.py
```diff
@@ -176,17 +178,18 @@ def _train_step(self, precision):
...
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, logits, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                    if self._is_finished(curr_step, end):
+                        return duration, self._finalize_periodic_logging(periodic)
```
abuccts: same

superbench/benchmarks/model_benchmarks/pytorch_gpt2.py
```diff
@@ -176,17 +178,18 @@ def _train_step(self, precision):
...
                     output = self._model(sample)
-                loss = self._loss_fn(output[range(self._args.batch_size), -1], self._target)
+                logits = output[range(self._args.batch_size), -1]
+                loss = self._loss_fn(logits.float(), self._target)
                 loss.backward()
                 self._optimizer.step()
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
```
abuccts, line 185: why remove this

superbench/benchmarks/model_benchmarks/pytorch_llama.py
```diff
@@ -196,17 +196,18 @@ def _train_step(self, precision):
...
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, logits, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                if self._is_finished(curr_step, end):
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: It's fine if we use the default frequency on 100 as the 2 aren't dependent, stop frequency can be different that the check_frequency parameter which focuses on recording the deterministic metrics only.

superbench/benchmarks/model_benchmarks/pytorch_lstm.py
```diff
@@ -148,17 +148,17 @@ def _train_step(self, precision):
...
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, output, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                if self._is_finished(curr_step, end):
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: It's fine if we use the default frequency on 100 as the 2 aren't dependent, stop frequency can be different that the check_frequency parameter which focuses on recording the deterministic metrics only.

superbench/benchmarks/model_benchmarks/pytorch_mixtral_impl.py
```diff
@@ -210,17 +235,18 @@ def _train_step(self, precision):
...
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, logits, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                if self._is_finished(curr_step, end):
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: It's fine if we use the default frequency on 100 as the 2 aren't dependent, stop frequency can be different that the check_frequency parameter which focuses on recording the deterministic metrics only.

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -182,17 +182,18 @@ def _train_step(self, precision):
...
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
-                    # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, logits, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                if self._is_finished(curr_step, end):
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: It's fine if we use the default frequency on 100 as the 2 aren't dependent, stop frequency can be different that the check_frequency parameter which focuses on recording the deterministic metrics only.

superbench/benchmarks/model_benchmarks/pytorch_gpt2.py
```diff
@@ -176,17 +176,19 @@ def _train_step(self, precision):
...
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
                     # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, logits, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                if self._is_finished(curr_step, end):
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: It's fine if we use the default frequency on 100 as the 2 aren't dependent, stop frequency can be different that the check_frequency parameter which focuses on recording the deterministic metrics only.

superbench/benchmarks/model_benchmarks/pytorch_cnn.py
```diff
@@ -108,17 +108,18 @@ def _train_step(self, precision):
...
                 end = self._timer()
                 curr_step += 1
                 if curr_step > self._args.num_warmup:
                     # Save the step time of every training/inference step, unit is millisecond.
                     duration.append((end - start) * 1000)
+                    self.record_determinism_fingerprint(curr_step, loss, output, periodic, self._args.check_frequency)
                     self._log_step_time(curr_step, precision, duration)
-                if self._is_finished(curr_step, end, check_frequency):
-                    return duration
+                if self._is_finished(curr_step, end):
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: It's fine if we use the default frequency on 100 as the 2 aren't dependent, stop frequency can be different that the check_frequency parameter which focuses on recording the deterministic metrics only.

superbench/benchmarks/model_benchmarks/pytorch_gpt2.py
```diff
@@ -176,17 +177,19 @@ def _train_step(self, precision):
...
+                    return duration, self._finalize_periodic_logging(periodic)
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: It's fine if we use the default frequency on 100 as the 2 aren't dependent, stop frequency can be different that the check_frequency parameter which focuses on recording the deterministic metrics only.

superbench/benchmarks/model_benchmarks/model_base.py
```diff
@@ -221,6 +232,12 @@ def _preprocess(self):
             self._result.set_return_code(ReturnCode.DISTRIBUTED_SETTING_INIT_FAILURE)
             return False
 
+        # Invoke model-specific deterministic seeding hook before dataset generation
+        try:
+            self.set_deterministic_seed()
+        except Exception:
+            logger.info('set_deterministic_seed() hook failed or not implemented for model: %s', self._name)
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: As per prior discussion, it was decided to not throw an exception here

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -30,15 +40,205 @@ def __init__(self, name, parameters=''):
...
+            required=False,
+            help='Random seed for deterministic training.',
+        )
+        self._parser.add_argument(
+            '--enable_determinism',
+            action='store_true',
+            default=False,
+            help='Enable deterministic training for reproducible results.',
+        )
+        self._parser.add_argument(
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

docs/user-tutorial/benchmarks/model-benchmarks.md
```diff
@@ -34,6 +34,34 @@ For inference, supported percentiles include
 
 **New: Support fp8_hybrid and fp8_e4m3 precision for BERT models.**
 
+**New: Deterministic Training Support**
+SuperBench now supports deterministic training to ensure reproducibility across runs. This includes fixed seeds and deterministic algorithms. To enable deterministic training, the following flags and environment variables must be set:
+
+- **Flags:**
+  - `--enable_determinism`: Enables deterministic computation for reproducible results.
+  - `--deterministic_seed <seed>`: Sets the seed for reproducibility.
+  - `--check_frequency <steps>`: How often to record deterministic metrics (default: 100).
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

superbench/benchmarks/model_benchmarks/pytorch_lstm.py
```diff
@@ -148,17 +148,17 @@ def _train_step(self, precision):
...
+                    return duration, self._finalize_periodic_logging(periodic)
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: not necessary

superbench/benchmarks/model_benchmarks/pytorch_mixtral_impl.py
```diff
@@ -210,17 +235,18 @@ def _train_step(self, precision):
...
+                    return duration, self._finalize_periodic_logging(periodic)
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: Not needed

superbench/benchmarks/model_benchmarks/pytorch_cnn.py
```diff
@@ -108,17 +108,18 @@ def _train_step(self, precision):
...
+                    return duration, self._finalize_periodic_logging(periodic)
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: not needed

superbench/benchmarks/model_benchmarks/pytorch_lstm.py
```diff
@@ -148,17 +148,17 @@ def _train_step(self, precision):
...
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: not needed

docs/user-tutorial/benchmarks/model-benchmarks.md
```diff
@@ -34,6 +34,34 @@ For inference, supported percentiles include
...
+
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

superbench/benchmarks/model_benchmarks/pytorch_mixtral_impl.py
```diff
@@ -195,8 +220,8 @@ def _train_step(self, precision):
             The step-time list of every training step.
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

superbench/benchmarks/model_benchmarks/pytorch_bert.py
```diff
@@ -167,8 +167,8 @@ def _train_step(self, precision):
             The step-time list of every training step.
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

superbench/benchmarks/model_benchmarks/pytorch_cnn.py
```diff
@@ -96,8 +96,8 @@ def _train_step(self, precision):
             The step-time list of every training step.
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

superbench/benchmarks/model_benchmarks/pytorch_lstm.py
```diff
@@ -136,8 +136,8 @@ def _train_step(self, precision):
             The step-time list of every training step.
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -30,15 +37,244 @@ def __init__(self, name, parameters=''):
...
+            check_frequency,
+            enable_determinism,
+            logger,
+        )
+
+    def _finalize_periodic_logging(self, periodic, info_key='loss'):
+        """Finalize periodic logging and return info dict for training step."""
+        info = {info_key: periodic.get(info_key, [])}
+        self._model_run_losses = list(periodic.get(info_key, []))
+        if self._model_run_periodic and getattr(self._args, 'enable_determinism', False):
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -30,15 +37,244 @@ def __init__(self, name, parameters=''):
...
+                    if self._global_rank is not None:
+                        metric_name = f'deterministic_{key}_rank{self._global_rank}'
+                    else:
+                        metric_name = f'deterministic_{key}'
+
+                    # Add summarized result (mean of checkpointed values)
+                    filtered_values = [v for v in values if v is not None]
+                    if filtered_values:
+                        self._result.add_result(metric_name, statistics.mean(filtered_values))
+                    else:
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: Stale

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -4,20 +4,27 @@
 """Module of the Pytorch model-benchmark base class."""
 
 import os
-from datetime import timedelta
+import random
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -30,15 +37,247 @@ def __init__(self, name, parameters=''):
...
+        to call torch.manual_seed().
+        """
+        if getattr(self._args, 'enable_determinism', False):
+            # Validate check_frequency before any deterministic operations
+            check_freq = getattr(self._args, 'check_frequency', 100)
+            if not isinstance(check_freq, int) or check_freq < 1:
+                logger.error(
+                    f'Invalid check_frequency={check_freq}. Must be a positive integer >= 1. '
+                    'Defaulting to 100.'
+                )
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

tests/benchmarks/model_benchmarks/test_pytorch_determinism_all.py
```diff
@@ -0,0 +1,155 @@
...
+    (
+        'gpt2-small',
+        '--batch_size 1 --num_classes 2 --seq_len 4 --num_warmup 1 --num_steps 20 '
+        '--model_action train --precision float32',
+    ),
+    pytest.param(
+        'llama2-7b',
+        '--batch_size 1 --seq_len 1 --num_warmup 1 --num_steps 20 --precision float32 --model_action train',
+        marks=pytest.mark.skip(
+            reason='Requires >26GB GPU memory for 7B model, and float16 incompatible with deterministic mode'
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

examples/benchmarks/pytorch_deterministic_example.py
```diff
@@ -0,0 +1,150 @@
...
+    '--model_action train --check_frequency 20',
+    'gpt2-small':
+    '--batch_size 1 --num_steps 300 --num_warmup 1 --seq_len 128 --precision float32 '
+    '--model_action train --check_frequency 20',
+    'llama2-7b':
+    '--batch_size 1 --num_steps 300 --num_warmup 1 --seq_len 512 --precision float32 --model_action train '
+    '--check_frequency 20',
+    'mixtral-8x7b':
+    '--hidden_size=4096 --num_hidden_layers=32 --num_attention_heads=32 --intermediate_size=14336 '
+    '--num_key_value_heads=8 --max_position_embeddings=32768 --router_aux_loss_coef=0.02 '
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: fixed

examples/benchmarks/pytorch_deterministic_example.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Convert to results-summary.jsonl format (flattened keys)
+    # Use format compatible with sb result commands: model-benchmarks:<category>/<benchmark>/<metric>
+    summary = {}
+    prefix = f'model-benchmarks:example:determinism/{benchmark_name}'
+    if 'result' in benchmark_results:
+        for metric, values in benchmark_results['result'].items():
+            # Use first value if it's a list
+            val = values[0] if isinstance(values, list) else values
+            # Add _rank0 suffix to deterministic metrics for compatibility with rules
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: it does not run on all ranks

superbench/analyzer/baseline_generation.py
```diff
@@ -174,19 +201,9 @@ def run(
...
-                    if isinstance(self._raw_data_df[metric].iloc[0], float):
-                        baseline[metric] = f'%.{digit}g' % val if abs(val) < 1 else f'%.{digit}f' % val
-                    elif isinstance(self._raw_data_df[metric].iloc[0], int):
-                        baseline[metric] = int(val)
-                    else:
-                        try:
-                            baseline[metric] = float(val)
-                        except Exception as e:
-                            logger.error('Analyzer: {} baseline is not numeric, msg: {}'.format(metric, str(e)))
+                baseline[metric] = self._format_metric_value(metric, baseline[metric], digit)
```
Aishwarya-Tonpe, line 204: Please note that I had to refactor this method due pipeline issue where the pipeline failed due to the method being flagged as too complex. Thank you.

superbench/analyzer/baseline_generation.py
```diff
@@ -150,6 +150,33 @@ def generate_baseline(self, algo, aggregated_df, diagnosis_rule_file, baseline):
...
+        if metric not in self._raw_data_df:
+            return val
+        sample = self._raw_data_df[metric].iloc[0]
+        if isinstance(sample, float):
+            # Keep full precision for deterministic metrics to avoid false positives in diagnosis
+            if 'deterministic' in metric:
+                return float(val)
+            return f'%.{digit}g' % val if abs(val) < 1 else f'%.{digit}f' % val
+        if isinstance(sample, int):
+            return int(val)
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: This is a part of code refactor, not an original change.

tests/benchmarks/model_benchmarks/test_pytorch_determinism_all.py
```diff
@@ -0,0 +1,146 @@
...
+from superbench.benchmarks import BenchmarkRegistry, Platform, Framework, ReturnCode
+
+# Set CUBLAS_WORKSPACE_CONFIG early to ensure deterministic cuBLAS behavior
+os.environ.setdefault('CUBLAS_WORKSPACE_CONFIG', ':4096:8')
+# Set PYTORCH_CUDA_ALLOC_CONF to avoid memory fragmentation
+os.environ.setdefault('PYTORCH_CUDA_ALLOC_CONF', 'expandable_segments:True')
+
+
+def run_deterministic_benchmark(model_name, params):
+    """Helper to launch a deterministic benchmark and return the result."""
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: Will take up incrementally

superbench/analyzer/baseline_generation.py
```diff
@@ -150,6 +150,33 @@ def generate_baseline(self, algo, aggregated_df, diagnosis_rule_file, baseline):
...
+        Returns:
+            The formatted metric value.
+        """
+        if metric not in self._raw_data_df:
+            return val
+        sample = self._raw_data_df[metric].iloc[0]
+        if isinstance(sample, float):
+            # Keep full precision for deterministic metrics to avoid false positives in diagnosis
+            if 'deterministic' in metric:
+                return float(val)
```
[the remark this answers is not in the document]
Aishwarya-Tonpe: Will take up incrementally

PR #733, opened 2025-09-05
Benchmarks: ib validation - add --set_ib_devices option to auto-select IB device by MPI local rank

yukirora:
**Description**
Summary:

**Major Revision**
- Add a new CLI flag --set_ib_devices to automatically select irregular IB devices based on the MPI local rank.
- When enabled, the benchmark queries available IB devices via network.get_ib_devices() and selects the device corresponding to OMPI_COMM_WORLD_LOCAL_RANK.
- Fall back to existing --ib_dev behavior when the flag is not provided.

**Minor Revision**
- Add an env in network.get_ib_devices() to allow user to set the device name

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
+++ b/superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
@@ -10,6 +10,7 @@ from superbench.common.utils import gen_pair_wise_config, gen_topo_aware_config
 from superbench.benchmarks import BenchmarkRegistry, ReturnCode
 from superbench.common.devices import GPU
 from superbench.benchmarks.micro_benchmarks import MicroBenchmarkWithInvoke
+from superbench.common.utils import network
 
 
 class IBBenchmark(MicroBenchmarkWithInvoke):
@@ -42,6 +43,12 @@ class IBBenchmark(MicroBenchmarkWithInvoke):
             default='mlx5_0',
             required=False,
             help='The IB device, e.g., mlx5_0, mlx5_$LOCAL_RANK, mlx5_$((LOCAL_RANK/2)), etc.',
+        )
+        self._parser.add_argument(
+            '--set_ib_devices',
+            action='store_true',
+            default=False,
+            help='Set irregular IB devices automatically according to the local rank when running with MPI. If IB devices are not able to scan, please use env IB_DEVICES to set them manually.',
         )
         self._parser.add_argument(
             '--gpu_dev',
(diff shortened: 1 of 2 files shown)


superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
```diff
@@ -43,6 +44,13 @@ def add_parser_arguments(self):
             required=False,
             help='The IB device, e.g., mlx5_0, mlx5_$LOCAL_RANK, mlx5_$((LOCAL_RANK/2)), etc.',
         )
+        self._parser.add_argument(
+            '--set_ib_devices',
+            action='store_true',
+            default=False,
+            help='Set irregular IB devices automatically according to the local rank. \
+            If IB devices are not able to scan, use env IB_DEVICES to set them manually.',
```
yzygitzh: to scan => to be probed?

superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
```diff
@@ -282,6 +290,16 @@ def __prepare_general_ib_command_params(self, msg_size, device='cpu'):
...
+        if self._args.set_ib_devices:
+            ib_devices = network.get_ib_devices()
+            local_rank = int(os.getenv('OMPI_COMM_WORLD_LOCAL_RANK', 0))
+            if local_rank >= len(ib_devices):
+                self._result.set_return_code(ReturnCode.INVALID_ARGUMENT)
+                logger.error(
+                    f'Local rank {local_rank} exceeds IB devices ({len(ib_devices)}) - benchmark: {self._name}'
+                )
+                return False
+            command_params = f'-F -n {self._args.iters} -d {ib_devices[local_rank].split(":")[0]} {msg_size} {gpu_dev}'
```
yzygitzh, line 302: Why we need `split(":")`

yukirora: the network.get_ib_devices will return like device:port

guoshzhao (approved):

superbench/common/utils/network.py
```diff
@@ -31,6 +32,8 @@ def get_ib_devices():
     Return:
         ib_devices_port (list): IB devices with available ports in current system.
     """
+    if os.getenv('IB_DEVICES', None):
+        return os.getenv('IB_DEVICES').split(',')
```
abuccts, line 36: this function returns a list of `ib:port`, not just a list of `ib`. seems this change is not compatible with previous function

superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
```diff
@@ -43,6 +44,13 @@ def add_parser_arguments(self):
             required=False,
             help='The IB device, e.g., mlx5_0, mlx5_$LOCAL_RANK, mlx5_$((LOCAL_RANK/2)), etc.',
         )
+        self._parser.add_argument(
+            '--set_ib_devices',
+            action='store_true',
+            default=False,
+            help='Set irregular IB devices automatically according to the local rank. \
+            If IB devices are not able to be probed, use env IB_DEVICES to set them manually.',
```
abuccts, line 52: if you set an env, it can be used in `--ib_dev` directly? e.g., `mlx5_$(echo "$IB_DEVICES" | cut -d, -f$((LOCAL_RANK + 1)))`

superbench/benchmarks/micro_benchmarks/ib_validation_performance.py
```diff
@@ -43,6 +44,13 @@ def add_parser_arguments(self):
             required=False,
             help='The IB device, e.g., mlx5_0, mlx5_$LOCAL_RANK, mlx5_$((LOCAL_RANK/2)), etc.',
         )
+        self._parser.add_argument(
+            '--set_ib_devices',
+            action='store_true',
+            default=False,
```
abuccts, line 50: or maybe change the type of this argument to `string` or `nargs="*"` with `default=None`
* default is `None`, keep previous behavior
* if set to `""` or `[]`, auto detect
* if set to `dev1,dev2` or `[dev1, dev2]`, use directly as the current `IB_DEVICES` env

superbench/common/utils/network.py
yukirora: we want to use the env to override this function(network.get_ib_devices) for both ib_validation and ib_loopback for some special use case like roce devices or gb200 bm, and IB_DEVICES env still can accept both ib1,ib2,ib3 or ib1:port1,ib2:port2,ib3:port3

PR #734, opened 2025-09-05
Benchmark: Model benchmark - add option to exclude data copy time in model benchmarks

yukirora:
**Description**
add option to exclude data copy time in model benchmarks.

**Major Revision**
- add an option --exclude_copy_time
- move start time after data copy finish in measured time

### DIFF
--- a/superbench/benchmarks/model_benchmarks/model_base.py
+++ b/superbench/benchmarks/model_benchmarks/model_base.py
@@ -151,7 +151,7 @@ class ModelBenchmark(Benchmark):
             '--no_copy',
             action='store_true',
             default=False,
-            help='Enable option to exclude data copy time.',
+            help='Exclude GPU data copy time from benchmark measurements.',
         )
 
     @abstractmethod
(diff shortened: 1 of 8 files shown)


guoshzhao (approved):

superbench/benchmarks/model_benchmarks/model_base.py
```diff
@@ -147,6 +147,13 @@ def add_parser_arguments(self):
             help='Real-time log every n steps.',
         )
 
+        self._parser.add_argument(
+            '--no_copy',
```
abuccts: exclude_copy_time? no_copy sounds like skip doing copy

superbench/benchmarks/model_benchmarks/model_base.py
```diff
@@ -147,6 +147,13 @@ def add_parser_arguments(self):
...
+            action='store_true',
+            default=False,
+            help='Enable option to exclude data copy time.',
```
abuccts: from gpt: Exclude data copy time from measured time.

superbench/benchmarks/model_benchmarks/model_base.py
yukirora: sure, updated

superbench/benchmarks/model_benchmarks/model_base.py
yukirora: updated

PR #736, opened 2025-09-08
Benchmarks: Micro benchmark - Add simultanneously all-to-host / host-to-all bandwidth testcases to nvbandwidth

yukirora:
**Description**
Add simultanneously all-to-host / host-to-all bandwidth testcases to nvbandwidth .

**Major Revision**
- nvbandwidth.patch: Add simultanneously all-to-host / host-to-all bandwidth testcases to nvbandwidth 
- upgrade nvbandwidth submodule into v0.8
- add patch into makefile build

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/dist_inference_cpp/dist_inference.cu
+++ b/superbench/benchmarks/micro_benchmarks/dist_inference_cpp/dist_inference.cu
@@ -416,8 +416,8 @@ void TestModel(int64_t m, int64_t n, int64_t k, float alpha, float beta, int32_t
     CHECK_CUBLASLT_ERROR(cublasLtMatrixLayoutCreate(&matF, CUDA_R_16F, k, n, k));
     CHECK_CUBLASLT_ERROR(cublasLtMatrixLayoutCreate(&matG, CUDA_R_16F, k, n, k));
 
-    CHECK_CUBLASLT_ERROR(cublasLtMatmulDescCreate(&matmul1, CUBLAS_COMPUTE_16F, CUDA_R_32F));
-    CHECK_CUBLASLT_ERROR(cublasLtMatmulDescCreate(&matmul2, CUBLAS_COMPUTE_16F, CUDA_R_32F));
+    CHECK_CUBLASLT_ERROR(cublasLtMatmulDescCreate(&matmul1, CUBLAS_COMPUTE_16F, CUDA_R_16F));
+    CHECK_CUBLASLT_ERROR(cublasLtMatmulDescCreate(&matmul2, CUBLAS_COMPUTE_16F, CUDA_R_16F));
 
     cublasOperation_t trans = CUBLAS_OP_N;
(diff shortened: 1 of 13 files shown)


yzygitzh:
A general question: why do we need to test fully parallel all-to-host and host-to-all?

yukirora:
> A general question: why do we need to test fully parallel all-to-host and host-to-all?

we want to check the nvidia c2c performance between grace gpu and grace cpu, which need fully parallel alltohost and sum, if we only use one gpu to the cpu, it's not accurate

third_party/nvbandwidth.patch
yzygitzh, line 1: Seems all in this patch are new content, and most of them are self-contained new classes and methods. 

Shall we move all in this patch to a separate new file, except for inline revision for `createTestcases` function and `Testcase` class? The new file can be copied to nvbandwidth project for building.

This should minimize the patch and make future development and maintenance easier.

third_party/nvbandwidth_testcases_patched.h
```diff
@@ -0,0 +1,121 @@
+// Copyright(c) Microsoft Corporation.
```
guoshzhao, line 1: Can we rename this file to `nvbandwidth_testcases_new.h`? Then in Makefile, we can execute `cp ../nvbandwidth_testcases_new.h ./testcases_new.h`.

yukirora: sure, updated

third_party/nvbandwidth.patch
yukirora: sure, moved the new classed and methods into  a new test_case header file

third_party/Makefile
```diff
@@ -299,5 +299,5 @@ endif
 
 # Build nvbandwidth.
 nvbandwidth: sb_micro_path
-	cd ./nvbandwidth && cmake . && make && cd ..
+	cd ./nvbandwidth && git apply ../nvbandwidth.patch && cp ../nvbandwidth_testcases_new.h ./ && mv ./nvbandwidth_testcases_new.h ./testcases_new.h && cmake . && make && cd ..
```
yzygitzh: why not `cp ../nvbandwidth_testcases_new.h ./testcases_new.h` directly

third_party/nvbandwidth.patch
```diff
@@ -0,0 +1,46 @@
+diff --git a/nvbandwidth.cpp b/nvbandwidth.cpp
+index 61a228f..488372a 100644
+--- a/nvbandwidth.cpp
++++ b/nvbandwidth.cpp
+@@ -29,6 +29,7 @@
+ #include "kernels.cuh"
+ #include "output.h"
+ #include "testcase.h"
++#include "testcases_new.h"
```
yzygitzh: `new` seems not accurate, how about `testcases_patched`

yzygitzh (approved):

third_party/Makefile
yukirora: updated, thanks

third_party/nvbandwidth.patch
yukirora: sure, updated

guoshzhao (approved):

PR #741, opened 2025-09-17
Benchmarks: Micro benchmark - Support verification and parallel run for disk performance benchmark

yzygitzh:
**Description**
Adds verification and parallel run support for disk performance benchmark.

**Major Revision**
- Adds `--verify` flag to support verify written data.
- Supports loading benchmark options from `PROC_RANK`, `BLOCK_DEVICES` and `NUMA_NODES` environmental variables.

### DIFF
--- a/.azure-pipelines/cpu-unit-test.yml
+++ b/.azure-pipelines/cpu-unit-test.yml
@@ -26,7 +26,7 @@ steps:
       echo "##vso[task.prependpath]$HOME/.local/bin"
     displayName: Export path
   - script: |
-      if [[ "$(echo $(imageTag) '< 3.11' | bc -l)" == 1 ]]; then
+      if python3 -c "import sys; exit(0 if sys.version_info < (3, 11) else 1)"; then
         python3 -m pip install --upgrade pip setuptools==65.7
       else
         python3 -m pip install --upgrade pip


superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -126,16 +126,58 @@ def add_parser_arguments(self):
...
-    def _preprocess(self):
+        self._parser.add_argument(
+            '--numa_node',
+            type=int,
+            required=False,
+            help='The index of NUMA node.',
+        )
+        self._parser.add_argument(
+            '--verify',
+            type=str,
```
yukirora, line 131: shall we add some choices?

superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -126,16 +126,58 @@ def add_parser_arguments(self):
...
             True if _preprocess() succeed.
         """
-        if not super()._preprocess():
+        if not super()._preprocess() or not self.__get_arguments_from_env():
             return False
 
-        fio_path = os.path.join(self._args.bin_dir, self._bin_name)
+        fio_basic_command = os.path.join(self._args.bin_dir, self._bin_name)
+        if self._args.numa_node is not None:
+            fio_basic_command = f'numactl -N {self._args.numa_node} {fio_basic_command}'
```
yukirora: just curious why need numactl -N for storage benchmarks, do we also need -C?

superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -126,16 +126,58 @@ def add_parser_arguments(self):
...
+            help='The index of NUMA node.',
+        )
+        self._parser.add_argument(
+            '--verify',
+            type=str,
+            required=False,
+            help='Verification method specified for fio --verify flag.',
+        )
+
+    def __get_arguments_from_env(self):
```
yukirora: I revised this into _get_arguments_from_env, and moved this and args.numa into MicroBenchmarkWithInvoke in another pr, so here you may can need to change the function name, replace the numa_node into numa

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: I expect this a pure wrapper of fio's `--verify`: https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-verify. It should abort with error when not meeting fio's expectation.

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: 1) Disks can be under different NUMA nodes, and we might need this to get max performance.
2) `-N` does node-level pinning while `-C` does core-level pinning. We only need the former granularity.

superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -126,16 +126,58 @@ def add_parser_arguments(self):
                     help='Number of threads in %s test.' % io_str,
                 )
 
-    def _preprocess(self):
+        self._parser.add_argument(
```
guoshzhao: In PR [#742], yuting will add one option `--numa` in base class, you can use it directly in this PR.

superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -126,16 +126,58 @@ def add_parser_arguments(self):
...
+        - BLOCK_DEVICES should be comma-separated strings, each being colon-separated block devices.
+          E.g., "/dev/nvme0n1:/dev/nvme1n1,/dev/nvme2n1:/dev/nvme3n1".
+        - NUMA_NODES should be comma-separated numbers. E.g., "0,1".
+        Note: The config from env variables will overwrite the configs defined in the command line.
+        """
+        try:
+            if os.getenv('PROC_RANK'):
+                rank = int(os.getenv('PROC_RANK'))
+                if os.getenv('BLOCK_DEVICES'):
+                    self._args.block_devices = os.getenv('BLOCK_DEVICES').split(',')[rank].split(':')
```
abuccts: why use env rather than argument, the parser is not robust

superbench/benchmarks/micro_benchmarks/disk_performance.py
abuccts: different benchmarks may have different numactl requirements, seems it's not unified

abuccts: different benchmarks may have different numactl requirements, seems it's not unified

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: fixed

superbench/benchmarks/micro_benchmarks/disk_performance.py
yukirora: can we add this link into 'help'? it can help user know which values should be set

yukirora (approved):

superbench/benchmarks/micro_benchmarks/disk_performance.py
```diff
@@ -126,16 +126,52 @@ def add_parser_arguments(self):
...
-        if not super()._preprocess():
+        if not super()._preprocess() or not self._get_arguments_from_env():
             return False
 
-        fio_path = os.path.join(self._args.bin_dir, self._bin_name)
+        fio_basic_command = os.path.join(self._args.bin_dir, self._bin_name)
+        if self._args.numa is not None:
+            fio_basic_command = f'numactl -N {self._args.numa} {fio_basic_command}'
+        if self._args.verify is not None:
+            fio_basic_command = f'{fio_basic_command} --verify={self._args.verify}'
```
yukirora, line 173: just curious if set verify, how to know the verify pass or not from the result?

yzygitzh: FIO will raise error message about which data blocks are corrupted, and it will be recorded in benchmark output. Also, the return code will be non-zero.

yukirora: got it, thanks

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: Added

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: fixed

guoshzhao: Could you please give examples of `--numa_node` in the `help` field?

guoshzhao (approved):

superbench/benchmarks/micro_benchmarks/disk_performance.py
yzygitzh: Change to ise block device indices in envs, block device strings will be passed from argument

PR #744, opened 2025-09-22
Enhancement: Add debug trace support

polarG:
To improve benchmark debugging, the following debug methods were added:

pytorch profiler in model benchmark

- SB_ENABLE_PYTORCH_PROFILER: switch to enable/disable
- SB_TORCH_PROFILER_TRACE_DIR: log path
These 2 runtime variables need to be configured in SB config file.

nsys in SB runner

- SB_ENABLE_NSYS: switch to enable/disable 
- SB_NSYS_TRACE_DIR: log path
These 2 runtime variables need to be configured in runner's ENV

### DIFF
--- a/superbench/benchmarks/model_benchmarks/pytorch_base.py
+++ b/superbench/benchmarks/model_benchmarks/pytorch_base.py
@@ -343,3 +343,58 @@ class PytorchBase(ModelBenchmark):
         if self._gpu_available:
             torch.cuda.synchronize()
         return time.time()
+
+
+    def _benchmark(self):
+        """Wrap super._benchmark with profiler context if enabled by environment variable.
+
+        Set SB_PYTORCH_PROFILER='true' or 1 to enable profiling.
+        """
+        # Check if this is not a CUDA GPU, if so call line 358
+        if not (torch.cuda.is_available() and torch.version.cuda is not None):
+            return super()._benchmark()
+
+        # Check if profiling is enabled via environment variable
+        enable_profiler = os.environ.get('SB_PYTORCH_PROFILER', '0').lower() in ('1', 'true')
+
+        if not enable_profiler:
+            # Run without profiling
+            return super()._benchmark()
+
+        # Run with profiling enabled
+        logger.info('PyTorch profiler enabled for model: {}'.format(self._name))
+        ret = None
+
+        from torch.profiler import profile, ProfilerActivity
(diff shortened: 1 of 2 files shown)


superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -343,3 +343,59 @@ def _timer(self):
         if self._gpu_available:
             torch.cuda.synchronize()
         return time.time()
+
+
+    def _benchmark(self):
+        """Wrap super._benchmark with profiler context if enabled by environment variable.
+
+        Set SB_PYTORCH_PROFILER='1' to enable profiling.
```
shcho: Minor suggestion to change SB_PYTORCH_PROFILER in the comment to SB_ENABLE_PYTORCH_PROFILER for consistency.

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -343,3 +343,58 @@ def _timer(self):
...
+        """
+        # Check if this is a Nvidia GPU
+        if not (torch.cuda.is_available() and torch.version.cuda is not None):
+            return super()._benchmark()
+
+        # Check if profiling is enabled via environment variable
+        enable_profiler = os.environ.get('SB_ENABLE_PYTORCH_PROFILER', '0').lower() == '1'
```
guoshzhao: Why call `lower()` here? and doesn't call for `enable_nsys = os.environ.get('SB_ENABLE_NSYS', '') == '1'`?

polarG: for lower, it's primarily ensuring that if someone accidentally sets SB_ENABLE_PYTORCH_PROFILER=1 (which would be a string), it gets properly compared.

nsys and pytorch profiler are 2 different tools. They will be enabled separately.

guoshzhao: What I mean is why you don't call `.lower()` for `enable_nsys = os.environ.get('SB_ENABLE_NSYS', '') == '1'`?

polarG: Yea, there is no diff for 1. I will delete it. Thanks!

polarG: Delete lower.

PR #750, opened 2025-10-09
Benchmarks: Micro benchmark - add nvbench based kernel-launch, sleep-kernel & auto-throughput

WenqingLan1:
This pull request adds support for NVBench-based GPU micro-benchmarks to SuperBench. 
- Integrated the NVBench submodule
- Implemented three benchmarks 
  - `nvbench-sleep-kernel`
  - `nvbench-kernel-launch`
  - `nvbench-auto-throughput`
- updated documentation and added example scripts

Example config:
```yaml
version: v0.12
superbench:
  enable:
  # nvbench benchmarks
  - nvbench-sleep-kernel:single
  - nvbench-sleep-kernel:list
  - nvbench-sleep-kernel:range
  - nvbench-sleep-kernel:range-step
  - nvbench-kernel-launch
  - nvbench-auto-throughput
  - nvbench-auto-throughput:stride-list
  - nvbench-auto-throughput:stride-range
  var:
    default_local_mode: &default_local_mode
      modes:
      - name: local
        proc_num: 4
        prefix: CUDA_VISIBLE_DEVICES={proc_rank}
        parallel: yes
  benchmarks:
    nvbench-sleep-kernel:single:
      <<: *default_local_mode
      timeout: 300
      parameters:
        duration_us: "50"                   # Single value format
        timeout: 30
    nvbench-sleep-kernel:list:
      <<: *default_local_mode
      timeout: 300
      parameters:
        duration_us: "[25,50,75]"         # List format - no spaces after commas
        timeout: 30
    nvbench-sleep-kernel:range:
      <<: *default_local_mode
      timeout: 300
      parameters:
        duration_us: "[0:5]"           # Range format
        timeout: 30
    nvbench-sleep-kernel:range-step:
      <<: *default_local_mode
      timeout: 300
      parameters:
        duration_us: "[0:50:10]"         # Range with step format
        timeout: 30
    nvbench-kernel-launch:
      <<: *default_local_mode
      timeout: 300
    nvbench-auto-throughput:
      <<: *default_local_mode
      timeout: 600
      parameters:
        stride: "[1,2,4,8]"              # List format for stride
        block_size: "[128,256,512,1024]"  # List format for block size
    nvbench-auto-throughput:stride-list:
      <<: *default_local_mode
      timeout: 600
      parameters:
        stride: "[1,2,4,8]"              # List format
        block_size: "[256,512]"
    nvbench-auto-throughput:stride-range:
      <<: *default_local_mode
      timeout: 600
      parameters:
        stride: "[1:8:2]"                # Range with step format
        block_size: "256"                 # Single value format
```


third_party/Makefile
```diff
@@ -311,3 +311,24 @@ endif
 nvbandwidth: sb_micro_path
 	cd ./nvbandwidth && git apply ../nvbandwidth.patch && cp ../nvbandwidth_testcases_patched.h ./testcases_patched.h && cmake . && make && cd ..
 	cp -v ./nvbandwidth/nvbandwidth $(SB_MICRO_PATH)/bin
+
+# Build nvbench
+cuda_nvbench: sb_micro_path
```
polarG: where is the nvbench checked out?

dockerfile/cuda13.0.dockerfile
```diff
@@ -62,6 +62,27 @@ RUN apt-get update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* /tmp/*
 
+# Install CMake 3.30.4 for nvbench compatibility
```
polarG: I checked the A100 base image, it's cmake version is 3.31.6. So we do not need to take care of this.

.gitmodules
```diff
@@ -33,3 +33,6 @@
 [submodule "third_party/nvbandwidth"]
 	path = third_party/nvbandwidth
 	url = https://github.com/NVIDIA/nvbandwidth.git
+[submodule "third_party/nvbench"]
+	path = third_party/nvbench
```
polarG, line 37: Do we need the specific release version?

superbench/benchmarks/micro_benchmarks/__init__.py
```diff
@@ -39,42 +39,18 @@
...
-    'RocmGemmFlopsBenchmark',
-    'RocmMemBwBenchmark',
-    'ShardingMatmul',
-    'TCPConnectivityBenchmark',
-    'TensorRTInferenceBenchmark',
-    'DirectXGPUEncodingLatency',
-    'DirectXGPUCopyBw',
-    'DirectXGPUMemBw',
-    'DirectXGPUCoreFlops',
-    'NvBandwidthBenchmark',
```
polarG, line 79: Any consideration here to change the format?

dockerfile/rocm5.0.x.dockerfile
```diff
@@ -98,7 +98,7 @@ RUN cd /tmp && \
 
 # Install Intel MLC
 RUN cd /tmp && \
-    wget -q https://downloadmirror.intel.com/793041/mlc_v3.11.tgz -O mlc.tgz && \
+    wget -q https://downloadmirror.intel.com/866182/mlc_v3.12.tgz -O mlc.tgz && \
```
polarG, line 101: Pls leave the mlc unchanged in this PR, i will create another pump PR to update the mlc version.

superbench/benchmarks/micro_benchmarks/__init__.py
WenqingLan1: it's done by lint.

.gitmodules
WenqingLan1: no, they don't have official release tags yet as a work-in-progress project.

dockerfile/cuda13.0.dockerfile
WenqingLan1: do you mean the based image of this dockerfile? I re-checked it and it has cmake version 3.31.6. will remove this.

dockerfile/cuda13.0.dockerfile
```diff
@@ -151,7 +151,7 @@ ADD dockerfile/etc /opt/microsoft/
 WORKDIR ${SB_HOME}
```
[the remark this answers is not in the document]
WenqingLan1: base image has cmake version 3.31.6

superbench/benchmarks/micro_benchmarks/nvbench_base.py
```diff
@@ -0,0 +1,232 @@
...
+import re
+from superbench.common.utils import logger
+from superbench.benchmarks import ReturnCode
+from superbench.benchmarks.micro_benchmarks.micro_base import MicroBenchmarkWithInvoke
+
+
+def parse_time_to_us(raw: str) -> float:
+    """Parse a time string like '123.45 us' or '1.5 s' to float microseconds."""
+    raw = raw.strip()
+    m = re.match(r'^([\d.]+)\s*([mun]?s)?$', raw)
```
polarG, line 20: **[SHOULD-FIX] `parse_time_to_us` does not accept the μ (micro sign) unit that callers' regexes expect**

```python
m = re.match(r'^([\d.]+)\s*([mun]?s)?$', raw)
```

The row-matching regexes in `nvbench_kernel_launch.py`, `nvbench_sleep_kernel.py`, and `nvbench_auto_throughput.py` all explicitly anticipate a `μ` (U+03BC MICRO SIGN) unit prefix, e.g.:

```python
r'([\d.]+\s*[μmun]?s)\s*\|\s*'    # CPU Time (μs, ns, ms, us, s)
```

but this helper's unit character class (`[mun]?s`) does not include `μ`. If the row regex ever captures a value like `"123.45 μs"`, passing it here raises `ValueError('Invalid time string: ...')`. That exception propagates out of the per-row parsing loop in each caller's `_process_raw_result`, is caught by the outer `except BaseException`, and triggers `_handle_parsing_error()` — marking the *entire* result as `MICROBENCHMARK_RESULT_PARSING_FAILURE` and discarding any rows already parsed in that call.

**Recommendation:** extend the regex to accept `μ` and normalize it before comparing, e.g.:
```python
m = re.match(r'^([\d.]+)\s*([μmun]?s)?$', raw)
...
unit = (m.group(2) or 'us').replace('μ', 'u')
```

.github/workflows/codeql-analysis.yml
```diff
@@ -52,7 +52,11 @@ jobs:
       - name: Install Dependency
         run: |
           DEBIAN_FRONTEND=noninteractive apt-get update
-          DEBIAN_FRONTEND=noninteractive apt-get install -y ffmpeg libavcodec-dev libavformat-dev libavutil-dev libswresample-dev sudo
+          DEBIAN_FRONTEND=noninteractive apt-get install -y ffmpeg libavcodec-dev libavformat-dev libavutil-dev libswresample-dev sudo build-essential
+      - name: Setup CMake
+        uses: lukka/get-cmake@latest
```
polarG, line 57: **[SHOULD-FIX] Third-party Action pinned to the floating `@latest` tag**

```yaml
uses: lukka/get-cmake@latest
```

`@latest` is a mutable ref — whoever controls `lukka/get-cmake` (or anyone who compromises that repo/release pipeline) can change what code runs in this CI job on every subsequent run, with no PR diff or review gate on this repo's side. This is a supply-chain compromise vector for the CodeQL analysis workflow, which typically runs with elevated permissions/secrets.

**Recommendation:** pin to an immutable commit SHA (preferred) or at minimum a specific release tag, e.g. `lukka/get-cmake@<full-commit-sha>` or `lukka/get-cmake@v3.30.5`.

dockerfile/cuda12.8.dockerfile
```diff
@@ -61,6 +61,27 @@ RUN apt-get update && \
...
+    cd /tmp && \
+    ARCH=$(uname -m) && \
+    case ${ARCH} in \
+        "aarch64") CMAKE_ARCH="aarch64" ;; \
+        "x86_64") CMAKE_ARCH="x86_64" ;; \
+        "arm64") CMAKE_ARCH="aarch64" ;; \
+        *) CMAKE_ARCH="x86_64" ;; \
+    esac && \
+    echo "Detected architecture: ${ARCH}, using CMAKE_ARCH: ${CMAKE_ARCH}" && \
+    wget -q https://github.com/Kitware/CMake/releases/download/v3.30.4/cmake-3.30.4-linux-${CMAKE_ARCH}.tar.gz && \
```
polarG, line 76: **[SHOULD-FIX] CMake release tarball downloaded and executed without integrity verification**

```dockerfile
wget -q https://github.com/Kitware/CMake/releases/download/v3.30.4/cmake-3.30.4-linux-${CMAKE_ARCH}.tar.gz && \
tar -xzf cmake-3.30.4-linux-${CMAKE_ARCH}.tar.gz && \
mv cmake-3.30.4-linux-${CMAKE_ARCH} /opt/cmake && \
ln -sf /opt/cmake/bin/* /usr/local/bin/ && \
```

The tarball is fetched over HTTPS but its checksum/signature is never verified before being extracted and symlinked onto `PATH`, where it's trusted for the rest of the image build (including compiling nvbench). A compromised release asset, mirror substitution, or interception on the download path would let arbitrary code execute during the image build with no detection (OWASP A08:2021 — Software and Data Integrity Failures).

**Recommendation:** verify a pinned SHA-256 checksum before extracting, e.g.:
```dockerfile
echo "<expected-sha256>  cmake-3.30.4-linux-${CMAKE_ARCH}.tar.gz" | sha256sum -c - && \
tar -xzf cmake-3.30.4-linux-${CMAKE_ARCH}.tar.gz && \
```
(Kitware publishes `cmake-3.30.4-SHA-256.txt` alongside each release.) This same block is duplicated in `dockerfile/cuda12.9.dockerfile` and needs the same fix there.

dockerfile/cuda12.9.dockerfile
```diff
@@ -62,6 +62,28 @@ RUN apt-get update && \
...
+    cd /tmp && \
+    ARCH=$(uname -m) && \
+    case ${ARCH} in \
+        "aarch64") CMAKE_ARCH="aarch64" ;; \
+        "x86_64") CMAKE_ARCH="x86_64" ;; \
+        "arm64") CMAKE_ARCH="aarch64" ;; \
+        *) CMAKE_ARCH="x86_64" ;; \
+    esac && \
+    echo "Detected architecture: ${ARCH}, using CMAKE_ARCH: ${CMAKE_ARCH}" && \
+    wget -q https://github.com/Kitware/CMake/releases/download/v3.30.4/cmake-3.30.4-linux-${CMAKE_ARCH}.tar.gz && \
```
polarG, line 77: **[SHOULD-FIX] CMake release tarball downloaded and executed without integrity verification**

Same issue as `dockerfile/cuda12.8.dockerfile`: the CMake tarball is downloaded over HTTPS but never checksum-verified before being extracted and symlinked onto `PATH`. See the comment on `cuda12.8.dockerfile` for the recommended `sha256sum -c` fix.

**[NON-BLOCKING] Also: this ~20-line CMake-install block is duplicated verbatim in both Dockerfiles.** Any future version bump or fix (like the checksum verification above) has to be applied in two places and can silently drift. Consider factoring it into a shared script (e.g. `dockerfile/scripts/install-cmake.sh`) that both Dockerfiles invoke.

polarG (changes requested):

.github/workflows/codeql-analysis.yml
WenqingLan1: should implement in a separate PR

.github/workflows/codeql-analysis.yml
```diff
@@ -52,7 +52,11 @@ jobs:
...
+        with:
+          cmakeVersion: '3.20.0'
```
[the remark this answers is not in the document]
WenqingLan1: should implement in a separate pr

superbench/benchmarks/micro_benchmarks/nvbench_base.py
```diff
@@ -0,0 +1,375 @@
...
+    if unit == 's':
+        return val * 1e6
+    elif unit == 'ns':
+        return val / 1e3
+    elif unit == 'ms':
+        return val * 1e3
+    return val
+
+
+_NVBENCH_INT_VALUES_PATTERN = re.compile(
```
polarG: To pass the lint check, pls use 1 line here.

polarG (approved):

PR #753, opened 2025-10-16
CI/CD - Fix python3.10 pipeline

polarG:
**Description**
Python3.10 pipeline failed.

**Solution**
From log, 'bc' cmd is missing. Since our image tags are simple, the solution is to remove 'bc' cmd directly.

### DIFF
--- a/.azure-pipelines/cpu-unit-test.yml
+++ b/.azure-pipelines/cpu-unit-test.yml
@@ -26,7 +26,7 @@ steps:
       echo "##vso[task.prependpath]$HOME/.local/bin"
     displayName: Export path
   - script: |
-      if [[ "$(imageTag)" < "3.11" ]]; then
+      if python3 -c "import sys; exit(0 if sys.version_info < (3, 11) else 1)"; then
         python3 -m pip install --upgrade pip setuptools==65.7
       else
         python3 -m pip install --upgrade pip


.azure-pipelines/cpu-unit-test.yml
```diff
@@ -26,7 +26,7 @@ steps:
       echo "##vso[task.prependpath]$HOME/.local/bin"
     displayName: Export path
   - script: |
-      if [[ "$(echo $(imageTag) '< 3.11' | bc -l)" == 1 ]]; then
+      if [[ "$(imageTag)" < "3.11" ]]; then
```
abuccts: what if imageTag is 3.7?

polarG: For python3.7, it will also use setuptools==65.7. 
In setup.py, the limitation is setuptools < 66 for py < 3.11.

guoshzhao (approved):

.azure-pipelines/cpu-unit-test.yml
abuccts: but your current condition won't work, [[ "3.7" < "3.11" ]] will be false so it won't install setuptools 65.7

polarG: Fixed.

abuccts (approved):

PR #754, opened 2025-10-31
CI/CD - Fix Azure test pipeline

polarG:
Python3.10 verification pipeline failed for conflict 'setuptools' version as below.
image

Root Cause:
The problem is that modern pip (25.3) uses an isolated build environment with the latest setuptools by default. The pipeline installs setuptools 65.7 in the user environment, but pip builds the package in an isolated environment with newer setuptools, which conflicts with the version check in [setup.py].

Solution:
Remove pip upgrade.

### DIFF
--- a/.azure-pipelines/cpu-unit-test.yml
+++ b/.azure-pipelines/cpu-unit-test.yml
@@ -27,11 +27,9 @@ steps:
     displayName: Export path
   - script: |
       if python3 -c "import sys; exit(0 if sys.version_info < (3, 11) else 1)"; then
-        python3 -m pip install --upgrade pip setuptools==65.7
-      else
-        python3 -m pip install --upgrade pip setuptools
+        python3 -m pip install --upgrade setuptools==65.7
       fi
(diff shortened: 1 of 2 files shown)


.azure-pipelines/cpu-unit-test.yml
```diff
@@ -29,9 +29,9 @@ steps:
       if python3 -c "import sys; exit(0 if sys.version_info < (3, 11) else 1)"; then
         python3 -m pip install --upgrade pip setuptools==65.7
       else
-        python3 -m pip install --upgrade pip
+        python3 -m pip install --upgrade pip setuptools
```
abuccts: why this is related to python3.10 pipeline failure?

polarG: After adding '--no-build-isolation' options, we are using the host py runtime env to build. In the host runtime env, there is no default setuptools installed.

abuccts:
> @abuccts I've opened a new pull request, #755, to work on those changes. Once the pull request is ready, I'll request review from you.

@copilot I mean in this PR, rather than opening a new one..

PR #769, opened 2025-12-19
Benchmark: Micro benchmark - Add float datatype support and other refinements to GPU Stream

WenqingLan1:
Refinements:
- Add support for float (fp32) execution and --data_type <float|double> CLI option for runtime type selection.
- Refactor CUDA kernels to use 128-bit vectorized accesses (double2 / float4) and move template kernel implementations into a header for cross-TU instantiation. (Required for CUDA template instantiation across compilation units.)
- Fix allocation buf size bug, args->size is buf size in bytes, not number of elements.
- Adjust execution/output to single visible GPU (device 0 via CUDA_VISIBLE_DEVICES) and update metric/tag formats (removing gpu_id) plus docs/examples/test log.
- Updated numa assignment from hard coded numa_alloc_onnode to numa_alloc_local to optimize memory allocation.
- Rename entry point file from gpu_stream_test.cpp to gpu_stream_main.cpp.

Note: metric tag removed gpu_idx and the execution is per-process, so users need to update the configs & rules.
New config:
```yaml
    gpu-stream:fp64:
      <<: *default_local_mode
      timeout: 600
      parameters:
        num_warm_up: 10
        num_loops: 40
        size: 4294967296
        data_type: double
    gpu-stream:fp64-correctness:
      <<: *default_local_mode
      timeout: 600
      parameters:
        num_warm_up: 0
        num_loops: 1
        size: 1048576
        data_type: double
        check_data: true
    gpu-stream:fp32:
      <<: *default_local_mode
      timeout: 600
      parameters:
        num_warm_up: 10
        num_loops: 40
        size: 2147483648
        data_type: float
    gpu-stream:fp32-correctness:
      <<: *default_local_mode
      timeout: 600
      parameters:
        num_warm_up: 0
        num_loops: 1
        size: 1048576
        data_type: float
        check_data: true
```

New rule:
```yaml
    gpu-stream:
      statistics:
        - mean
      categories: GPU-STREAM
      aggregate: True
      metrics:
        - gpu-stream:fp(?:32|64)/STREAM_.*_(?:bw|ratio):(\d+)
```

Example results:
```json
"gpu-stream:fp32/STREAM_COPY_float_buffer_2617245696_block_256_bw:0": 1234, 
"gpu-stream:fp32/STREAM_COPY_float_buffer_2617245696_block_256_bw:1": 1234, 
"gpu-stream:fp32/STREAM_COPY_float_buffer_2617245696_block_256_bw:2": 1234, 
"gpu-stream:fp32/STREAM_COPY_float_buffer_2617245696_block_256_bw:3": 1234
```
Processed by rules:
```md
| gpu-stream:fp32/STREAM_COPY_float_buffer_2617245696_block_256_bw | mean | 1234|
```


superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -631,21 +632,29 @@ int GpuStream::Run() {
...
+    // Run on CUDA device 0 (the visible GPU assigned by CUDA_VISIBLE_DEVICES).
+    if (opts_.data_type == "float") {
+        auto args = std::make_unique<BenchArgs<float>>();
+        args->gpu_id = 0;
+        cudaGetDeviceProperties(&args->gpu_device_prop, 0);
+        args->num_warm_up = opts_.num_warm_up;
+        args->num_loops = opts_.num_loops;
+        args->size = opts_.size;
+        args->check_data = opts_.check_data;
+        bench_args_.emplace_back(std::move(args));
```
guoshzhao: Can we make it a template function?

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -668,14 +677,6 @@ int GpuStream::Run() {
                 // Print device info with both the memory clock and peak bandwidth
                 PrintCudaDeviceInfo(curr_args->gpu_id, curr_args->gpu_device_prop, memory_clock_mhz, peak_bw);
 
-                // Set the NUMA node
-                ret = numa_run_on_node(curr_args->numa_id);
-                if (ret != 0) {
```
guoshzhao, line 673: numa_alloc_local allocates on the current NUMA node of the calling thread at allocation time, which is correct. But if the OS later migrates the thread to a remote NUMA node, remote memory accesses introduce latency. Should we pin the thread to the fix NUMA node?

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream_kernels.hpp
```diff
@@ -7,13 +7,195 @@
...
+ *
+ * @tparam T The type of the value to fetch.
+ * @param[out] v The register to write the fetched value to.
+ * @param[in] p The source memory location to fetch the value from.
+ */
+template <typename T> inline __device__ void Fetch(T &v, const T *p) {
+#if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)
+    v = *p;
+#else
+    if constexpr (std::is_same<T, float>::value) {
```
guoshzhao: Should handle the "unsupported type" here and other similar places?

guoshzhao (changes requested):

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream_kernels.hpp
WenqingLan1: The type is constrained at the entry point. VectorType<T> only has specializations for float and double, so any unsupported type would fail to compile with "incomplete type error on VecT<T>" error. No runtime guard is needed.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
WenqingLan1: added back the thread pinning logic

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
WenqingLan1: fixed

guoshzhao (approved):

polarG (dismissed):
## Multi-model code review — PR #769

Synthesized from three subagent perspectives (Claude / Gemini / OpenAI) by the
`code-review` orchestrator. Each finding is grounded in code at HEAD
`5cec42c7`. Severities: **BLOCKER / SHOULD-FIX / NON-BLOCKING / NOTED**.

### Summary

| # | Severity | Title | Location | Agreement |
|---|---|---|---|---|
| 1 | SHOULD-FIX | Unchecked `numa_alloc_local` return → nullptr deref | `gpu_stream.cu` L240 / L247 | Claude + OpenAI (flagged BLOCKER; downgraded — pre-existing pattern but PR touches the lines) |
| 2 | SHOULD-FIX | Unchecked `cudaGetDeviceProperties` result | `gpu_stream.cu` L618 | Claude + OpenAI (flagged BLOCKER; downgraded — same rationale) |
| 3 | SHOULD-FIX | Thread pinned to host CPU's NUMA node, not the GPU's affinity node | `gpu_stream.cu` L660-L674 | OpenAI |
| 4 | SHOULD-FIX | Metric tag format break (`_gpu_<id>_` removed) — no baseline migration noted | docs / `gpu_stream.cu` L590 | Synthesizer (cross-cuts the PR) |
| 5 | SHOULD-FIX | Hard-coded `kBytesPerThread = 16` decoupled from `VecT<T>` | `gpu_stream.cu` L426 | Gemini |
| 6 | SHOULD-FIX | Float code path is not exercised end-to-end (no float log, no float parsing test) | `tests/data/gpu_stream.log` / `test_gpu_stream.py` | Claude + OpenAI |
| 7 | SHOULD-FIX | Renamed `gpu_stream_main.cpp` docstring is stale (no `--data_type`) | `gpu_stream_main.cpp` L7-L14 | Gemini |
| 8 | NON-BLOCKING | `memcmp` validation of float arrays is fragile (works only because current test values are exactly representable) | `gpu_stream.cu` L355 | Claude (flagged BLOCKER; downgraded after FP analysis) |
| 9 | NON-BLOCKING | String-based `data_type` dispatch + silent ternary fallback to `double` | `gpu_stream.cu` L658 | Gemini + OpenAI |
| 10 | NON-BLOCKING | Per-component `if constexpr` block duplicated across Scale/Add/Triad | `gpu_stream_kernels.hpp` L118-L188 | Gemini |
| 11 | NON-BLOCKING | `std::variant` over a single active alternative is over-engineered | `gpu_stream.hpp` L36 | Gemini |
| 12 | NON-BLOCKING | Missing defensive guard for `num_threads_per_block == 0` (currently safe via constexpr array) | `gpu_stream.cu` L427 | Claude |
| 13 | NOTED | `kNumLoopUnroll` removed — vectorized 128-bit ops should compensate, but please confirm bandwidth on A100/H100 | `gpu_stream_kernels.hpp` | Gemini + OpenAI |
| 14 | NOTED | Single-GPU per-process execution is intentional under `default_local_mode` (matches `CUDA_VISIBLE_DEVICES={proc_rank}`) | `gpu_stream.cu` L654 | OpenAI |
| 15 | NOTED | With default `--size = 4 GiB`, `default_local_mode` × 8 procs × `--check_data` ≈ 160 GiB host RAM; new configs in PR description avoid this by using ~1.2 GiB | `gpu_stream.cu` L238 | OpenAI |

### Cross-cutting observations (no inline anchor)

- The PR description references new YAML entries (`gpu-stream:fp64`, `gpu-stream:fp32`, etc.) and a new analyzer rule, but **no `superbench/config/*.yaml` is updated in this PR**. Please clarify whether a follow-up PR will add them, otherwise the new metric tag format will not be aggregated by `result_summary`.
- The new metric format `STREAM_<op>_<dtype>_buffer_<size>_block_<block>` removes `gpu_<id>`. **Existing baselines, dashboards, and rules that reference the old name will break silently.** Consider mentioning this in the release notes / CHANGELOG.

### Verdict

**NEEDS-CHANGES** — no true blockers, but several SHOULD-FIX items (unchecked
allocator returns, NUMA-affinity mismatch with the GPU, missing float test
coverage, stale docstring, and a hard-coded constant decoupled from the kernel
type) should be addressed before merge.

> Reviewed by the `code-review` agent (orchestrator) at
> `/home/hongtaozhang/repos/msv/agents/code-review/code-review.md`.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -235,15 +236,15 @@ template <typename T> int GpuStream::PrepareBufAndStream(std::unique_ptr<BenchAr
     cudaError_t cuda_err = cudaSuccess;
 
     if (args->check_data) {
-        // Generate data to copy
-        args->sub.data_buf = static_cast<T *>(numa_alloc_onnode(args->size * sizeof(T), args->numa_id));
+        // Generate data to copy - use local NUMA node for best CPU access
+        args->sub.data_buf = static_cast<T *>(numa_alloc_local(args->size));
```
polarG, line 245: **[SHOULD-FIX] Unchecked `numa_alloc_local` return → nullptr deref**

`numa_alloc_local()` returns `nullptr` on failure (insufficient memory, NUMA disabled, etc.). The next loop dereferences `args->sub.data_buf[j]` without a null check — and the same is true for `check_buf` on L247. Since this PR is editing these lines (switching from `numa_alloc_onnode`), it's a good chance to add the check.

```cpp
args->sub.data_buf = static_cast<T *>(numa_alloc_local(args->size));
if (args->sub.data_buf == nullptr) {
    std::cerr << "PrepareBufAndStream::numa_alloc_local data_buf failed" << std::endl;
    return -1;
}
// ... and after the second alloc:
if (args->sub.check_buf == nullptr) {
    std::cerr << "PrepareBufAndStream::numa_alloc_local check_buf failed" << std::endl;
    numa_free(args->sub.data_buf, args->size);
    return -1;
}
```

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -605,12 +607,28 @@ int GpuStream::RunStream(std::unique_ptr<BenchArgs<T>> &args, const std::string
...
 
+/**
+ * @brief Creates and initializes a BenchArgs for the given type and adds it to bench_args_.
+ *
+ * @tparam T The data type (float or double) for the benchmark arguments.
+ */
+template <typename T> void GpuStream::CreateBenchArgs() {
+    auto args = std::make_unique<BenchArgs<T>>();
+    args->gpu_id = 0;
+    cudaGetDeviceProperties(&args->gpu_device_prop, 0);
```
polarG: **[SHOULD-FIX] `cudaGetDeviceProperties` return value not checked**

If `cudaGetDeviceProperties` fails, `args->gpu_device_prop` is left uninitialized; subsequent reads of `memoryBusWidth`, `ECCEnabled`, `name`, etc. in `PrintCudaDeviceInfo` and the peak-bandwidth math then operate on garbage.

```cpp
cudaError_t cuda_err = cudaGetDeviceProperties(&args->gpu_device_prop, 0);
if (cuda_err != cudaSuccess) {
    std::cerr << "CreateBenchArgs::cudaGetDeviceProperties error: " << cuda_err << std::endl;
    return; // and propagate failure (e.g., via a return-code or by leaving variant empty)
}
```

The `void` return type makes propagation awkward — consider returning `int` so `Run()` can bail out cleanly.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -631,71 +649,64 @@ int GpuStream::Run() {
...
+    if (cpu < 0) {
+        std::cerr << "Run::sched_getcpu failed" << std::endl;
         return -1;
     }
+    int local_node = numa_node_of_cpu(cpu);
+    if (local_node < 0) {
+        std::cerr << "Run::numa_node_of_cpu failed for cpu " << cpu << std::endl;
+        return -1;
+    }
+    if (numa_run_on_node(local_node) != 0) {
```
polarG: **[SHOULD-FIX] NUMA pinning targets the host process's current node, not the GPU's affinity node**

`sched_getcpu()` + `numa_node_of_cpu()` returns wherever the kernel happened to schedule this process. Under `default_local_mode` (8 parallel processes, one per `CUDA_VISIBLE_DEVICES={proc_rank}`), the OS may pin all 8 processes to the same socket initially, even when the assigned GPU lives on the other socket. The result is that `numa_alloc_local` allocates on a node that is *remote* to the GPU — exactly the host↔device path measured by `--check_data` (`cudaMemcpy(..., cudaMemcpyDefault)`).

Recommend pinning to the **GPU's** affinity node instead:

```cpp
// Query GPU 0's preferred CPU NUMA node (e.g., via NVML)
int gpu_numa = -1;
nvmlDevice_t dev;
if (nvmlInit() == NVML_SUCCESS &&
    nvmlDeviceGetHandleByIndex(0, &dev) == NVML_SUCCESS) {
    nvmlDeviceGetNumaNodeId(dev, &gpu_numa);
    nvmlShutdown();
}
int target = (gpu_numa >= 0) ? gpu_numa : numa_node_of_cpu(sched_getcpu());
if (numa_run_on_node(target) != 0) { ... }
```

Also note: this `numa_run_on_node` happens *after* `cudaGetDeviceProperties` (which can implicitly init the CUDA context). Migrating threads after CUDA init may cause TLB / socket-affinity surprises.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -583,11 +586,10 @@ int GpuStream::RunStream(std::unique_ptr<BenchArgs<T>> &args, const std::string
 
     // output formatted results to stdout
     // Tags are of format:
-    // STREAM_<Kernelname>_datatype_gpu_<gpu_id>_buffer_<buffer_size>_block_<block_size>
-    for (int i = 0; i < args->sub.times_in_ms.size(); i++) {
-        std::string tag = "STREAM_" + KernelToString(i) + "_" + data_type + "_gpu_" + std::to_string(args->gpu_id) +
-                          "_buffer_" + std::to_string(args->size);
-        for (int j = 0; j < args->sub.times_in_ms[i].size(); j++) {
+    // STREAM_<Kernelname>_datatype_buffer_<buffer_size>_block_<block_size>
+    for (size_t i = 0; i < args->sub.times_in_ms.size(); i++) {
```
polarG, line 643: **[SHOULD-FIX] Metric tag format is a breaking change — flag it explicitly**

The output format dropped `_gpu_<id>` from every STREAM tag (and switched to TAB separation):

- old: `STREAM_COPY_double_gpu_0_buffer_<size>_block_<block>`
- new: `STREAM_COPY_double_buffer_<size>_block_<block>`

This silently breaks any downstream consumers — existing baseline JSONs, `result_summary` rules, dashboards, and external scripts that key on the old metric name. The PR description includes a *new* aggregation rule for the new format, but neither the diff nor the docs include a deprecation note or migration guidance.

Please:
1. Call out the format change in the PR description / release notes / CHANGELOG.
2. Update or add the YAML rule (the `gpu-stream:` aggregation) referenced in the PR description — it isn't in the diff today.
3. Confirm no in-repo baselines reference the old format (none were found in `superbench/config/*.yaml`, but please check internal/example baselines).

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -420,10 +421,12 @@ int GpuStream::RunStreamKernel(std::unique_ptr<BenchArgs<T>> &args, Kernel kerne
     int size_factor = 2;
 
     // Validate data size
-    uint64_t num_elements_in_thread_block = kNumLoopUnroll * num_threads_per_block;
-    uint64_t num_bytes_in_thread_block = num_elements_in_thread_block * sizeof(T);
+    // Each thread processes 128 bits (16 bytes) for optimal memory bandwidth.
+    // For double: uses double2 (16 bytes). For float: would use float4 (16 bytes).
+    constexpr uint64_t kBytesPerThread = 16; // 128-bit aligned access
```
polarG: **[SHOULD-FIX] `kBytesPerThread = 16` is decoupled from `VecT<T>`**

The host-side block-size math hard-codes 16 bytes per thread, while the actual vector type is decided in `gpu_stream_kernels.hpp` via `VectorType<T>`. If anyone changes `double2 → double4`, or `float4 → float2`, or adds a new dtype, this constant silently goes wrong and `num_thread_blocks = args->size / num_bytes_in_thread_block` will produce a mismatched grid — at best a launch failure, at worst silent OOB.

Recommend deriving from the same source of truth:

```cpp
constexpr uint64_t kBytesPerThread = sizeof(VecT<T>);
static_assert(kBytesPerThread == 16, "Vector type must be 128-bit aligned for current PTX");
```

This gives both compile-time safety and a single point of change.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream_kernels.hpp
```diff
@@ -7,13 +7,187 @@
...
+template <typename T> __global__ void ScaleKernel(VecT<T> *tgt, const VecT<T> *src, const T scalar) {
+    uint64_t index = blockIdx.x * blockDim.x + threadIdx.x;
+    VecT<T> val;
+    Fetch(val, src + index);
+    if constexpr (std::is_same<T, double>::value) {
+        val.x *= scalar;
+        val.y *= scalar;
+    } else if constexpr (std::is_same<T, float>::value) {
+        val.x *= scalar;
+        val.y *= scalar;
```
polarG, line 126: **[NON-BLOCKING] Repeated `if constexpr` per-component blocks**

Every Scale/Add/Triad kernel does the same `if constexpr (double) { .x .y } else if constexpr (float) { .x .y .z .w }` dance — three copies of the same dispatch, easy to drift if a new dtype (e.g. `__half2`/`bfloat162`) is added later. Consider a small per-component helper:

```cpp
template <typename Vec, typename Op>
__device__ inline void ForEachComponent(Vec& v, Op op) {
    op(v.x); op(v.y);
    if constexpr (requires { v.z; }) { op(v.z); op(v.w); }
}
```

Or simply iterate via `reinterpret_cast<T*>(&val)` and `sizeof(VecT<T>)/sizeof(T)`. Either way you get one source of truth.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.hpp
```diff
@@ -34,8 +34,8 @@ class GpuStream {
     int Run();
 
   private:
```
polarG, line 36: **[NON-BLOCKING] `std::variant` for a single active alternative**

`bench_args_` is now exactly one of float/double — never both, never empty during use. The variant adds a discriminator and forces every consumer through `std::visit` + `if constexpr`. Two simpler alternatives:

1. Make `GpuStream` a class template parameterized by `T` and select the instantiation in `main()` after parsing `--data_type`.
2. Keep a `std::unique_ptr<IBenchRunner>` to a small abstract base whose two derived classes are `BenchRunner<float>` / `BenchRunner<double>`.

Both eliminate the `std::visit` + `if constexpr` cascade in `Run()`.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -631,71 +649,64 @@ int GpuStream::Run() {
...
-            variant_args);
-    }
-    if (has_error) {
+    if (gpu_count < 1) {
+        std::cerr << "Run::No GPU available" << std::endl;
+        return -1;
+    }
+
+    // Run on CUDA device 0 (the visible GPU assigned by CUDA_VISIBLE_DEVICES).
+    opts_.data_type == "float" ? CreateBenchArgs<float>() : CreateBenchArgs<double>();
```
polarG: **[NON-BLOCKING] String-based dispatch + silent ternary fallback**

```cpp
opts_.data_type == "float" ? CreateBenchArgs<float>() : CreateBenchArgs<double>();
```

If `ParseOpts`'s validation is ever bypassed (or someone adds a new dtype but forgets this site), any unknown string silently falls through to `double`. Suggestions:

1. Replace the string with `enum class DataType { kDouble, kFloat };` so the compiler enforces exhaustive handling at every dispatch site.
2. Or at minimum, make the dispatch explicit:

```cpp
if (opts_.data_type == "float") {
    CreateBenchArgs<float>();
} else if (opts_.data_type == "double") {
    CreateBenchArgs<double>();
} else {
    std::cerr << "Run::Invalid data_type: " << opts_.data_type << std::endl;
    return -1;
}
```

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -420,10 +421,12 @@ int GpuStream::RunStreamKernel(std::unique_ptr<BenchArgs<T>> &args, Kernel kerne
     int size_factor = 2;
 
     // Validate data size
-    uint64_t num_elements_in_thread_block = kNumLoopUnroll * num_threads_per_block;
-    uint64_t num_bytes_in_thread_block = num_elements_in_thread_block * sizeof(T);
+    // Each thread processes 128 bits (16 bytes) for optimal memory bandwidth.
+    // For double: uses double2 (16 bytes). For float: would use float4 (16 bytes).
+    constexpr uint64_t kBytesPerThread = 16; // 128-bit aligned access
+    uint64_t num_bytes_in_thread_block = num_threads_per_block * kBytesPerThread;
```
polarG, line 462: **[NON-BLOCKING] No defensive guard against `num_threads_per_block == 0`**

Currently safe — `num_threads_per_block` only ever comes from the constexpr `kThreadsPerBlock = {128, 256, 512, 1024}` array. But if that ever becomes runtime-driven, both the modulo at L431 and the divide at L432 become UB. A one-liner now is cheap insurance:

```cpp
if (num_threads_per_block == 0) {
    std::cerr << "RunStreamKernel::num_threads_per_block must be > 0" << std::endl;
    return -1;
}
```

tests/data/gpu_stream.log
```diff
@@ -2,40 +2,23 @@ STREAM Benchmark
 Buffer size(bytes): 4294967296
 Number of warm up runs: 10
 Number of loops: 40
```
polarG, line 4: **[SHOULD-FIX] Float code path is not exercised end-to-end**

The new fp32 path is added in C++ and CLI, but the test data here only contains `Data type: double`. As a result:

- `_test_gpu_stream_result_parsing` only validates parsing of `STREAM_*_double_*` lines; the float tag format (`STREAM_*_float_*`) is never exercised.
- `test_gpu_stream_command_generation_cuda_float` only verifies that `--data_type float` appears in the command string — it does not verify that the binary or parser actually handles a float result.

Recommend either:
1. Adding a second log fixture (e.g. `tests/data/gpu_stream_float.log`) plus a parametrised parsing test, **or**
2. Appending a few `STREAM_*_float_*` lines to this file and asserting the parser emits both `_float_*_bw` and `_float_*_ratio` keys.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream_kernels.hpp
```diff
@@ -7,13 +7,187 @@
...
+ * to the target array. This is used to measure transfer rates without any arithmetic operations.
+ *
+ * @param[out] tgt The target array where data will be copied to (128-bit aligned).
+ * @param[in] src The source array from which data will be copied (128-bit aligned).
+ */
+template <typename T> __global__ void CopyKernel(VecT<T> *tgt, const VecT<T> *src) {
+    uint64_t index = blockIdx.x * blockDim.x + threadIdx.x;
+    VecT<T> val;
+    Fetch(val, src + index);
+    Store(tgt + index, val);
```
polarG, line 104: **[NOTED] Loop-unroll (`kNumLoopUnroll = 2`) removed in favor of a single 128-bit vector op**

Each thread now performs exactly one `Fetch` + one `Store` (vs. two unrolled scalar pairs before). On bandwidth-bound STREAM workloads this is usually a wash or a slight win because:

- 128-bit `ld.volatile.global.v2.f64` / `v4.f32` matches the L2 → L1 sector size on modern GPUs.
- `volatile` qualifiers are preserved, so the compiler still cannot coalesce or cache.

However, please **measure on representative GPUs** (V100 / A100 / H100 and MI250 / MI300 if HIP support is reactivated later) before merge to confirm no regression on the latency-bound side (e.g., consumer GPUs without HBM).

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -631,71 +649,64 @@ int GpuStream::Run() {
...
-                    has_error = true;
-                }
-            },
-            variant_args);
-    }
-    if (has_error) {
+    if (gpu_count < 1) {
+        std::cerr << "Run::No GPU available" << std::endl;
+        return -1;
+    }
```
polarG, line 713: **[NOTED] Single-GPU execution by design — relies on `default_local_mode`**

Multi-GPU iteration was removed; the binary now only benchmarks `cudaSetDevice(0)`. Under `default_local_mode` (`prefix: CUDA_VISIBLE_DEVICES={proc_rank}`, `proc_num: 8`), this is correct — Superbench fans out one process per GPU. The change does, however:

- Break standalone usage (`./bin/gpu_stream` no longer measures every GPU on the host).
- Couple the binary to Superbench's process model (now reflected in the function comment).

This is intentional per the PR description; the only ask is to call this behavior change out in the user-tutorial doc (`docs/user-tutorial/benchmarks/micro-benchmarks.md`) so users running the binary directly are not surprised.

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream_kernels.hpp
```diff
@@ -7,13 +7,187 @@
...
+    uint64_t index = blockIdx.x * blockDim.x + threadIdx.x;
+    VecT<T> val_a;
+    VecT<T> val_b;
+    Fetch(val_a, src_a + index);
+    Fetch(val_b, src_b + index);
+    if constexpr (std::is_same<T, double>::value) {
+        val_b.x += (val_a.x * scalar);
+        val_b.y += (val_a.y * scalar);
+    } else if constexpr (std::is_same<T, float>::value) {
+        val_b.x += (val_a.x * scalar);
```
polarG, line 186: **[NON-BLOCKING] `memcmp` validation of float results is fragile**

`CheckBuf` (in `gpu_stream.cu`, unchanged but newly reachable for `T = float`) compares result buffers byte-for-byte via `memcmp`. For TRIAD specifically, the GPU may compile `val_b.x += val_a.x * scalar` to a single FMA (one rounding) while the host validation buffer in `PrepareValidationBuf` is computed with two separate roundings.

For the **current** test values (`j % 256 ∈ [0, 255]`, `scalar = 11.0`) every intermediate is exactly representable in IEEE-754 single precision, so `memcmp` *does* agree on this PR. But this is a fragile invariant — increasing `kUInt8Mod` or changing `scalar` could silently break `--check_data` for fp32 only.

Recommend, when convenient, switching to a tolerance-based comparison for `float`:

```cpp
constexpr float kRelTol = 1e-6f;
for (size_t i = 0; i < n; ++i) {
    if (std::abs(actual[i] - expected[i]) > kRelTol * std::abs(expected[i])) ...
}
```

superbench/benchmarks/micro_benchmarks/gpu_stream/gpu_stream.cu
```diff
@@ -235,15 +236,15 @@ template <typename T> int GpuStream::PrepareBufAndStream(std::unique_ptr<BenchAr
     cudaError_t cuda_err = cudaSuccess;
 
     if (args->check_data) {
```
polarG, line 243: **[NOTED] Host-RAM cost when `--check_data` is on**

With the default `--size = 4 GiB`, every process allocates `data_buf + check_buf = 8 GiB` of host memory plus three GPU buffers. Under `default_local_mode` × 8 processes × `--check_data`, that's ~64 GiB of host RAM just for STREAM validation — enough to OOM smaller dev nodes.

The new YAML configs in the PR description avoid this by overriding to `~1.2 GiB`, so this is **not** a blocker, just a watch-out: please consider documenting the per-process memory cost of `--check_data` next to the `--size` flag in the user tutorial.

polarG (approved):

PR #773, opened 2026-01-16
Benchmark: Micro benchmark - Fix precision loss in kernel launch overhead benchmark

alephpiece:
**Description**

The original code uses integer division in the time calculation: (end_tv.tv_usec) / 1000. Since an empty kernel launch overhead is typically very small (often less than 1ms), this integer division truncates the microsecond component to 0, leading to significant precision loss and incorrect results.

image

**Major Revision**
- Changed the divisor from 1000 to 1000. (floating-point literal). This forces floating-point arithmetic, preserving the microsecond precision required for accurate overhead measurement.

### DIFF
--- a/superbench/benchmarks/micro_benchmarks/kernel_launch_overhead/kernel_launch.cu
+++ b/superbench/benchmarks/micro_benchmarks/kernel_launch_overhead/kernel_launch.cu
@@ -60,8 +60,7 @@ double test_cuda_kernel_launch_wall_time(int num_warmups, int num_steps) {
         EmptyKernel<<<1, 1>>>();
         cudaDeviceSynchronize();
         gettimeofday(&end_tv, NULL);
-        total_time += (end_tv.tv_sec - begin_tv.tv_sec) * 1000.0 +
-                      (end_tv.tv_usec - begin_tv.tv_usec) / 1000.0;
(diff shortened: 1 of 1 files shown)


polarG (approved):

guoshzhao (approved):

guoshzhao:
@alephpiece could you please run `python3 setup.py format` and `python3 setup.py lint` to fix the lint check?

alephpiece:
@guoshzhao 
I may be missing something, but I’m not sure why `python3 setup.py format` and `python3 setup.py lint` would help for this PR. The change here is only a one-line fix in a `.cu` file, and there are no Python file changes. Also, `CPP code lint` is already passing.

guoshzhao:
> @guoshzhao I may be missing something, but I’m not sure why `python3 setup.py format` and `python3 setup.py lint` would help for this PR. The change here is only a one-line fix in a `.cu` file, and there are no Python file changes. Also, `CPP code lint` is already passing.

Oh, sorry I didn't look into it carefully, the python lint failed due to some pipeline issue not your changes. We will fix it. Sorry for the confusion.
image

  File "/__w/1/s/.eggs/setuptools_scm-10.0.5-py3.12.egg/setuptools_scm/__init__.py", line 8, in <module>
    from vcs_versioning import Configuration
ModuleNotFoundError: No module named 'vcs_versioning'

PR #796, opened 2026-03-26
Bugfix - Fix DeepSpeed BF16 config validation error

polarG:
**Description**
 The megatron-gpt:deepspeed benchmark fails with return code 3 (INVALID_BENCHMARK_RESULT) during the BF16 training round. The benchmark runs two precision   rounds (FP16 then BF16), and while FP16 succeeds, BF16 crashes at DeepSpeed initialization with:
  pydantic_core._pydantic_core.ValidationError: 5 validation errors for DeepSpeedBF16Config
     loss_scale          - Extra inputs are not permitted
     loss_scale_window   - Extra inputs are not permitted
     min_loss_scale      - Extra inputs are not permitted
     initial_scale_power - Extra inputs are not permitted
     hysteresis          - Extra inputs are not permitted

**Root Cause**
__prepare_deespeed_config() in megatron_gpt3.py uses the same precision_template for both FP16 and BF16 configs. This template includes loss-scaling parameters (loss_scale, loss_scale_window, min_loss_scale, initial_scale_power, hysteresis) that are valid for FP16 but rejected by DeepSpeedBF16Config, which uses pydantic strict validation and does not accept extra fields.

BF16 does not need loss scaling because it has sufficient dynamic range to avoid the underflow/overflow issues that FP16 faces.

This was always technically incorrect, but only became a hard failure when DeepSpeed migrated from Pydantic v1 to Pydantic v2 (around DeepSpeed v0.15–v0.16). In Pydantic v1, the extra = "forbid" setting was less strictly enforced, so the extra fields were silently ignored. Pydantic v2 strictly rejects all unknown fields with a ValidationError.

**Fix**
 Generate precision-specific DeepSpeed configs:

   - FP16: includes all loss-scaling parameters (unchanged behavior)
   - BF16: only {'enabled': True}
   - FP32: no precision section

  This fix is backward compatible — passing only {'enabled': True} for BF16 is valid in all DeepSpeed versions, since the loss-scaling fields were never used by BF16.


superbench/benchmarks/model_benchmarks/megatron_gpt3.py
```diff
@@ -307,15 +307,23 @@ def __prepare_deespeed_config(self, precision_megatron):
         """Prepare deepspeed configs."""
         self._config_json_path = os.path.join(self._args.data_home, 'ds_config_gpt.json')
         # Load deepspeed config template json file
```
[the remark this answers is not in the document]
polarG: Addressed in d3baec91: updated the comment to "Build deepspeed config template in memory" since the template is constructed inline in this function rather than loaded from a JSON file.

guoshzhao (approved):

tests/benchmarks/model_benchmarks/test_megatron_gpt.py
```diff
@@ -500,6 +500,69 @@ def test_deepseek_v2_command(self):
...
+        benchmark._MegatronGPT__prepare_deespeed_config('bf16')
+        with open(benchmark._config_json_path) as f:
+            cfg = json.load(f)
+        self.assertEqual(cfg.get('bf16'), {'enabled': True})
+        self.assertNotIn('fp16', cfg)
+        for forbidden in ('loss_scale', 'loss_scale_window', 'min_loss_scale', 'initial_scale_power', 'hysteresis'):
+            self.assertNotIn(forbidden, cfg['bf16'])
+
+        # fp16: keeps loss-scale fields
+        benchmark._MegatronGPT__prepare_deespeed_config('fp16')
```
WenqingLan1: The fp16 test constructs the benchmark without --hysteresis, so the `if self._args.hysteresis is not None` branch is never exercised. Worth adding the param and assertion to cover that branch.

superbench/benchmarks/model_benchmarks/megatron_gpt3.py
```diff
@@ -306,16 +306,24 @@ def _parse_log(self, output):
     def __prepare_deespeed_config(self, precision_megatron):
```
WenqingLan1: typo deespeed -> deepspeed.

WenqingLan1 (approved):

superbench/benchmarks/model_benchmarks/megatron_gpt3.py
polarG: Addressed in 607a7a21: renamed `__prepare_deespeed_config` to `__prepare_deepspeed_config` (definition and the single internal call site in `_megatron_command`).

tests/benchmarks/model_benchmarks/test_megatron_gpt.py
```diff
@@ -500,6 +500,69 @@ def test_deepseek_v2_command(self):
...
+            ),
+        )
+        mock_generate_dataset.return_value = True
+        assert benchmark._preprocess() is True
+
+        # bf16: only {'enabled': True}, no loss-scale fields, no fp16 section
+        benchmark._MegatronGPT__prepare_deespeed_config('bf16')
+        with open(benchmark._config_json_path) as f:
+            cfg = json.load(f)
+        self.assertEqual(cfg.get('bf16'), {'enabled': True})
```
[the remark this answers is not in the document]
polarG: Addressed in 607a7a21: the test no longer invokes the name-mangled private method directly. `test_megatron_gpt_deepspeed_config` now calls the public-shaped `benchmark._megatron_command(Precision.BFLOAT16 | FLOAT16 | FLOAT32)` and validates the resulting DeepSpeed config file written by that path (BF16 omits the loss-scale knobs; FP16 round-trips `hysteresis`; FP32 writes neither `fp16` nor `bf16` sections). This removes coupling to the private implementation name.

tests/benchmarks/model_benchmarks/test_megatron_gpt.py
polarG: Addressed in 607a7a21: the FP16 case in `test_megatron_gpt_deepspeed_config` now constructs the benchmark with `--hysteresis 3` and asserts `cfg['fp16']['hysteresis'] == 3`, exercising the `if self._args.hysteresis is not None` branch.

superbench/benchmarks/model_benchmarks/megatron_gpt3.py
```diff
@@ -328,7 +336,7 @@ def __prepare_deespeed_config(self, precision_megatron):
             'prescale_gradients': self._args.prescale_grad,
         }
 
-        if len(precision_megatron) > 0:
+        if precision_template is not None:
             ds_config_template[precision_megatron] = precision_template
```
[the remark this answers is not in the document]
polarG: Addressed in 607a7a21: `test_megatron_gpt_deepspeed_config` now exercises BF16 end-to-end via `benchmark._megatron_command(Precision.BFLOAT16)` and asserts the generated DeepSpeed config contains `{'bf16': {'enabled': True}}` while none of `loss_scale`, `loss_scale_window`, `min_loss_scale`, `initial_scale_power`, `hysteresis` appear in the BF16 config — pinning the BF16 schema change against the upstream DeepSpeed validator.

PR #805, opened 2026-04-17
CI/CD - Fix setuptools-scm 10.x compatibility for Python 3.12

polarG:
## Description
This affects all PRs running `python3 setup.py lint` on the Python 3.12 CI job.

## Root Cause

Comparing the last successful cpu-unit-test build (58939, Mar 25) with a recent failing build (58996, Apr 14), the Python 3.12 "Install dependencies" step shows:

| Package | Successful (Mar 25) | Failing (Apr 14) |
|---|---|---|
| `setuptools-scm` | < 10.0 (no `vcs-versioning` dep) | 10.0.5 (requires `vcs-versioning`) |

`setuptools-scm` 10.0.5 was released between the two runs and added `vcs-versioning` as a new dependency. The `setup_requires` mechanism in `setup.py` does not install transitive dependencies, so `vcs-versioning` is missing at runtime.

The successful build lint log (Python 3.12): "ModuleNotFoundError: No module named 'vcs_versioning'"

This affects all PRs running `python3 setup.py lint` on the Python 3.12 CI job.

## Changes

- Add `vcs_versioning` explicitly to `setup_requires` in `setup.py` so it is available when `setuptools-scm` is imported during `setup.py` execution.

## Testing

Verified that `setuptools-scm` 10.0.5 declares `vcs-versioning` as a dependency, and the CI failure matches the missing transitive dependency pattern.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,8 +15,9 @@ jobs:
       uses: actions/checkout@v2
     - name: Install dependencies
       run: |
-        curl -L -o /tmp/install-misspell.sh https://raw.githubusercontent.com/client9/misspell/master/install-misspell.sh
-        sudo bash /tmp/install-misspell.sh -b /bin
+        curl -fsSL --retry 3 -o /tmp/misspell.tar.gz https://github.com/client9/misspell/releases/download/v0.3.4/misspell_0.3.4_linux_64bit.tar.gz
+        tar -xzf /tmp/misspell.tar.gz -C /tmp
+        sudo install /tmp/misspell /bin/misspell
     - name: Check spelling
       run: |
         misspell -error .
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,8 @@ jobs:
       uses: actions/checkout@v2
     - name: Install dependencies
       run: |
-        curl -L https://git.io/misspell | sudo bash -s -- -b /bin
+        curl -L -o /tmp/install-misspell.sh https://raw.githubusercontent.com/client9/misspell/master/install-misspell.sh
+        sudo bash /tmp/install-misspell.sh -b /bin
     - name: Check spelling
       run: |
         misspell -error .
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,14 +13,12 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
-    - name: Install dependencies
+    - name: Install misspell
       run: |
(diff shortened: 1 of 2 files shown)


guoshzhao:
@polarG Looks Cuda UT and Spelling check also have some issues, pls take a look:
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.2") 
-- Checking for module 'libavcodec'
--   No package 'libavcodec' found

setup.py
```diff
@@ -162,6 +162,7 @@ def run(self):
     },
     setup_requires=[
         'setuptools_scm',
```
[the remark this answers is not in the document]
polarG: We have anchored the missing dependency and would like to match the version by python version automatically.

.github/workflows/lint.yml
```diff
@@ -15,7 +15,9 @@ jobs:
       uses: actions/checkout@v2
     - name: Install dependencies
       run: |
-        curl -L https://git.io/misspell | sudo bash -s -- -b /bin
+        curl -fsSL --retry 3 -o /tmp/misspell.tar.gz https://github.com/client9/misspell/releases/download/v0.3.4/misspell_0.3.4_linux_64bit.tar.gz
```
guoshzhao, line 18: Why the original way doesn't work? And can we use the package management tool to install it as copilot suggested?

guoshzhao (approved):

.github/workflows/lint.yml
polarG: The original `curl -L https://git.io/misspell | sudo bash -s -- -b /bin` failed because:

1. **git.io is deprecated** — GitHub deprecated the URL shortener in 2022. Existing redirects are kept in a read-only archive but the service is unmaintained, leading to intermittent failures.

2. **curl-to-bash without `-f` is fragile** — The install script involves multiple chained network calls (git.io redirect → raw install script → GitHub release lookup → asset download). Without `--fail`, curl silently passes HTTP error pages to bash, causing cryptic failures.

3. **The failure was transient** — Lint runs on April 18 using the same old config succeeded, confirming intermittency rather than a permanent breakage. We could not pull the exact CI logs (403) to identify which specific network hop failed.

Switched to `go install github.com/client9/misspell/cmd/misspell@v0.3.4` as suggested. This uses Go's checksum database (sumdb) for integrity verification, requires no sudo, and Go is pre-installed on ubuntu-latest runners.

WenqingLan1 (approved):

PR #807, opened 2026-04-21
Config: Update config files

alephpiece:


alephpiece:
Sorry, I accidentally opened this PR against the upstream repository while using `gh pr create`; I’ve closed it and recreated the PR against the correct target branch in my fork.

PR #810, opened 2026-05-01
Add ROCm 6.4, 7.0 and 7.2 support

arsdragonfly:


polarG:
@arsdragonfly Thanks for your contribution, could we break down this PR to 3 separate PR? Then each PR contain 1 ROCM version. Thanks!

PR #817, opened 2026-05-19
Enhancement: Add rocprofv2 trace support for AMD GPUs

shcho:
**Description**
Extend trace generation to support AMD GPUs using rocprofv2

- runner.py: Add SB_ENABLE_ROCPROF/SB_ROCPROF_TRACE_DIR env vars to enable rocprofv2 profiling (--hip-trace --kernel-trace --plugin json) in local, torch.distributed, and mpi modes
- pytorch_base.py: Extend GPU guard to support ROCm (torch.version.hip) so PyTorch profiler works on AMD GPUs


superbench/runner/runner.py
```diff
@@ -135,12 +135,23 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+            if enable_nsys and mode.proc_rank == 0:
+                trace_command = (
+                    f'nsys profile --output {trace_dir}/{benchmark_name}_{mode.proc_rank}_traces '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof and mode.proc_rank == 0:
+                trace_command = (
+                    f'rocprofv2 --hip-trace --kernel-trace --plugin json '
+                    f'-d {rocprof_trace_dir}/{benchmark_name}_{mode.proc_rank}_traces '
+                )
```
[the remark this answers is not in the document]
shcho: Addressed.

superbench/runner/runner.py
```diff
@@ -135,12 +135,23 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
         enable_nsys = os.environ.get('SB_ENABLE_NSYS', '') == '1'
         trace_dir = os.environ.get('SB_NSYS_TRACE_DIR', self._sb_output_dir)
 
+        # Enable rocprofv2 profiling based on environment variable
+        enable_rocprof = os.environ.get('SB_ENABLE_ROCPROF', '') == '1'
+        rocprof_trace_dir = os.environ.get('SB_ROCPROF_TRACE_DIR', self._sb_output_dir)
```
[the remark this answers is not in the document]
shcho: Addressed.

superbench/runner/runner.py
```diff
@@ -159,10 +170,17 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof:
+                nsys_prefix = (
+                    f'rocprofv2 --hip-trace --kernel-trace --plugin json '
+                    f'-d {rocprof_trace_dir}/{benchmark_name}_traces '
+                )
 
             mode_command = (
                 f'{nsys_prefix}'
```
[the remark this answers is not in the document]
shcho: Addressed.

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -591,8 +591,8 @@ def _benchmark(self):
         Run the benchmark then handle post-run model log save/compare.
         Set SB_ENABLE_PYTORCH_PROFILER='1' to enable profiling.
         """
-        # Check if this is a Nvidia GPU
-        if not (torch.cuda.is_available() and torch.version.cuda is not None):
+        # Check if this is a Nvidia or AMD GPU
+        if not (torch.cuda.is_available() and (torch.version.cuda is not None or torch.version.hip is not None)):
```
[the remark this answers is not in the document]
shcho: Same mechanism as related PR.

superbench/runner/runner.py
```diff
@@ -159,10 +170,17 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+            nsys_prefix = ''
+            if enable_nsys:
+                nsys_prefix = (
+                    f'nsys profile --output {trace_dir}/{benchmark_name}_traces '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof:
+                nsys_prefix = (
+                    f'rocprofv2 --hip-trace --kernel-trace --plugin json '
+                    f'-d {rocprof_trace_dir}/{benchmark_name}_traces '
```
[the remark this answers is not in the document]
shcho: rocprof uses PID as trace file name. No overwritten risk.

superbench/runner/runner.py
```diff
@@ -159,23 +173,41 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+                trace_prefix = (
+                    f'nsys profile --output {trace_output} '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof:
+                trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_traces')
+                trace_prefix = (
+                    f'rocprofv2 --hip-trace --kernel-trace --plugin json '
+                    f'-d {trace_output} '
+                )
```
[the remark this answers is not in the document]
shcho: Low priority. Intended to preserve same code structure as previous similar PR.

superbench/runner/runner.py
```diff
@@ -135,12 +136,25 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+                trace_command = (
+                    f'nsys profile --output {trace_output} '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof and mode.proc_rank == 0:
+                trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
+                trace_command = (
+                    f'rocprofv2 --hip-trace --kernel-trace --plugin json '
+                    f'-d {trace_output} '
+                )
```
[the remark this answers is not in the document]
shcho: The trace options are internal options and users are expected to only select either NSYS or RPCPROF to use.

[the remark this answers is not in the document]
shcho: Addressed in commit.

superbench/benchmarks/model_benchmarks/pytorch_base.py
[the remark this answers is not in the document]
shcho: Addressed in commit.

shcho:
@microsoft-github-policy-service agree company="Microsoft"

superbench/runner/runner.py
```diff
@@ -135,12 +136,25 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
-            ) if enable_nsys and mode.proc_rank == 0 else ''
+            trace_command = ''
+            if enable_nsys and mode.proc_rank == 0:
+                trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
+                trace_command = (
+                    f'nsys profile --output {trace_output} '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof and mode.proc_rank == 0:
+                trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
```
polarG: **[BLOCKER] `shlex.quote()` output is re-embedded inside an already single-quoted shell string, defeating the escaping**

**Evidence:**
```python
trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
trace_command = (
    f'rocprofv2 --hip-trace --kernel-trace --plugin json '
    f'-d {trace_output} -- '
)
```
and, downstream (unchanged by this PR, in `_run_proc`):
```python
fcmd = "docker exec {env_list} sb-workspace bash -lc '{command}'"
...
ansible_runner_config = self._ansible_client.get_shell_config(
    fcmd.format(env_list=env_list, command=self.__get_mode_command(benchmark_name, mode, timeout))
)
```

**Issue:** `shlex.quote()` only produces a safe token when it becomes the *outermost* shell's literal input. Here the quoted `trace_output` is spliced into `mode_command`, which is itself substituted into `'{command}'` — a slot that is **already wrapped in literal single quotes** by `fcmd` (`bash -lc '{command}'`). When `SB_ROCPROF_TRACE_DIR`/`SB_NSYS_TRACE_DIR` contains a character that forces `shlex.quote` to add its own single quotes (spaces, `$`, backticks, etc.), the resulting `'...'` sequence prematurely closes the outer quoting, leaving the interior content **unquoted** before the next `'` reopens a new segment. This applies to both the new rocprof path and the refactored nsys path, in all three modes (`local`, `torch.distributed`, `mpi`).

**Impact:** Anything that can set `SB_ROCPROF_TRACE_DIR`/`SB_NSYS_TRACE_DIR` (job config env, CI variables, a shared/templated environment) can inject shell syntax (e.g. `$(...)` command substitution) that executes with the privileges of the process running the Ansible shell module — potentially root — on every target node.

**Recommendation:** Don't quote a fragment that will be spliced into another literal-quoted string. Either (a) pass the trace directory through Ansible as a separate variable/argument instead of string concatenation, (b) validate/allowlist `SB_NSYS_TRACE_DIR`/`SB_ROCPROF_TRACE_DIR` against a safe path pattern before use, or (c) quote the entire inner command once, at the point it is embedded into `fcmd`, rather than quoting sub-fragments earlier. Add a regression test with a trace dir containing a space and a `$(...)` sequence, asserting the final shell string stays safe.

superbench/runner/runner.py
```diff
@@ -135,12 +136,25 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
-                f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
-            ) if enable_nsys and mode.proc_rank == 0 else ''
+            trace_command = ''
+            if enable_nsys and mode.proc_rank == 0:
+                trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
+                trace_command = (
+                    f'nsys profile --output {trace_output} '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof and mode.proc_rank == 0:
```
polarG: **[SHOULD-FIX] Silent, undocumented precedence when both `SB_ENABLE_NSYS` and `SB_ENABLE_ROCPROF` are set**

**Evidence:**
```python
if enable_nsys and mode.proc_rank == 0:
    ...
elif enable_rocprof and mode.proc_rank == 0:
    ...
```

**Issue:** The two profilers are mutually exclusive via `if`/`elif` in all three modes, but there is no validation, warning, or log message when both flags are set simultaneously — `SB_ENABLE_ROCPROF` is silently dropped whenever `SB_ENABLE_NSYS=1` is also set.

**Impact:** A user targeting an AMD node with `SB_ENABLE_NSYS` left set (e.g. from a shared/global env or copy-pasted config) gets no rocprofv2 output and no indication why, since `nsys` doesn't work on non-NVIDIA hardware — a confusing silent no-op.

**Recommendation:** Log a warning when both are enabled and document the precedence, or raise a config validation error in `__validate_sb_config`. Add a test that sets both env vars and asserts the intended behavior.

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -591,8 +591,10 @@ def _benchmark(self):
         Run the benchmark then handle post-run model log save/compare.
         Set SB_ENABLE_PYTORCH_PROFILER='1' to enable profiling.
         """
-        # Check if this is a Nvidia GPU
-        if not (torch.cuda.is_available() and torch.version.cuda is not None):
+        # Check if this is a Nvidia or AMD GPU
+        if not (torch.cuda.is_available() and (
+            getattr(torch.version, 'cuda', None) is not None or getattr(torch.version, 'hip', None) is not None
```
polarG: **[SHOULD-FIX] GPU guard broadened to admit ROCm, but downstream CUDA-specific profiler internals are unverified for HIP**

**Evidence:**
```python
if not (torch.cuda.is_available() and (
    getattr(torch.version, 'cuda', None) is not None or getattr(torch.version, 'hip', None) is not None
)):
```
...followed by (unchanged, but now reachable on ROCm when `SB_ENABLE_PYTORCH_PROFILER=1`):
```python
diag_agent_prof = profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA], record_shapes=True)
...
if event.device_type != DeviceType.CPU:
    continue
...
diag_agent_event['gpu_time'] = event.cuda_time
```

**Issue:** This guard change is the only thing deciding whether AMD GPUs now enter the profiler branch, but the profiling logic itself (`ProfilerActivity.CUDA`, `event.cuda_time`, `event.device_type` checks) was not updated or verified against ROCm-build PyTorch semantics.

**Impact:** On AMD nodes with `SB_ENABLE_PYTORCH_PROFILER=1`, this could raise an error or silently produce empty/misleading `gpu_time` data for a code path that previously never executed on AMD.

**Recommendation:** Verify (or add CI coverage for) `SB_ENABLE_PYTORCH_PROFILER=1` on a ROCm build before merging; fail loudly rather than silently if expected profiler attributes are absent.

superbench/runner/runner.py
```diff
@@ -159,23 +173,41 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+                trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_traces')
+                trace_prefix = (
+                    f'nsys profile --output {trace_output} '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof:
+                trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_traces')
+                trace_prefix = (
+                    f'rocprofv2 --hip-trace --kernel-trace --plugin json '
+                    f'-d {trace_output} -- '
```
polarG: **[SHOULD-FIX] rocprofv2 wraps the launcher process, not the per-rank worker processes, in `torch.distributed`/`mpi` modes**

**Evidence:**
```python
elif enable_rocprof:
    trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_traces')
    trace_prefix = (
        f'rocprofv2 --hip-trace --kernel-trace --plugin json '
        f'-d {trace_output} -- '
    )
mode_command = (
    f'{trace_prefix}'
    f'torchrun'
    f' --no_python --nproc_per_node={mode.proc_num} {torch_dist_params}{exec_command}'
    ...
)
```

**Issue:** `rocprofv2` is invoked as a prefix to `torchrun`/`mpirun` themselves, rather than per spawned worker process. Unlike `nsys`, whose ability to trace descendant processes is well established, ROCProfiler's tracing model for children forked/spawned by a wrapped parent is not verified here.

**Impact:** For multi-process modes — the primary GPU-benchmarking use case — the resulting trace may reflect only the launcher process and miss per-rank GPU kernel activity, silently producing an empty or misleading trace while appearing to succeed.

**Recommendation:** Verify on real AMD hardware that rocprofv2 captures traces for all ranks spawned under `torchrun`/`mpirun`; if not, invoke it per-rank (similar to how `local` mode profiles `mode.proc_rank == 0` directly).

superbench/runner/runner.py
```diff
@@ -159,23 +173,41 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
-            ) if enable_nsys else ''
+            trace_command = ''
+            if enable_nsys:
+                trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
+                trace_command = (
+                    f'nsys profile --output {trace_output} '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof:
+                trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
```
polarG: **[SHOULD-FIX] No test coverage for the new `enable_rocprof` command-construction branches**

**Evidence:**
```python
elif enable_rocprof:
    trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
    trace_command = (
        f'rocprofv2 --hip-trace --kernel-trace --plugin json '
        f'-d {trace_output} -- '
    )
```

**Issue:** Three new branches (`local`, `torch.distributed`, `mpi`) mirror the existing nsys behavior, but no test changes accompany the PR. A parametrized test asserting the exact generated command string per mode — including a trace dir containing a space — is the natural place a quoting regression like the one flagged above would be caught.

**Impact:** Regressions in command construction (including the injection issue noted separately, or future refactors) will not be caught by CI.

**Recommendation:** Add parametrized unit tests for `__get_mode_command` covering `enable_rocprof=True` for all three modes plus a trace dir with whitespace/special characters, and a case with both `SB_ENABLE_NSYS`/`SB_ENABLE_ROCPROF` set.

superbench/benchmarks/model_benchmarks/pytorch_base.py
```diff
@@ -591,8 +591,10 @@ def _benchmark(self):
         Run the benchmark then handle post-run model log save/compare.
         Set SB_ENABLE_PYTORCH_PROFILER='1' to enable profiling.
         """
-        # Check if this is a Nvidia GPU
-        if not (torch.cuda.is_available() and torch.version.cuda is not None):
+        # Check if this is a Nvidia or AMD GPU
+        if not (torch.cuda.is_available() and (
```
polarG: **[SHOULD-FIX] No test coverage for the widened GPU-guard branch matrix**

**Evidence:**
```python
if not (torch.cuda.is_available() and (
    getattr(torch.version, 'cuda', None) is not None or getattr(torch.version, 'hip', None) is not None
)):
```

**Issue:** No test mocks `torch.version.cuda`/`torch.version.hip`/`torch.cuda.is_available` to verify the four combinations (cuda-only, hip-only, neither, both) route to the correct branch.

**Impact:** Regressions in this guard (e.g. accidentally reverting to CUDA-only, or breaking the CPU-only skip path) would not be caught by CI, especially where ROCm hardware isn't available in CI.

**Recommendation:** Add a unit test using `unittest.mock.patch` on `torch.version`/`torch.cuda.is_available` covering all four combinations.

superbench/runner/runner.py
```diff
@@ -159,23 +173,41 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
                 '--nnodes=$NNODES --node_rank=$NODE_RANK --master_addr=$MASTER_ADDR --master_port=$MASTER_PORT '
             )
 
-            nsys_prefix = (
-                f'nsys profile --output {trace_dir}/{benchmark_name}_traces '
-                f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
-            ) if enable_nsys else ''
+            trace_prefix = ''
```
polarG: **[SHOULD-FIX] Trace-command construction logic tripled across `local`/`torch.distributed`/`mpi` branches**

**Evidence:**
```python
# local
if enable_nsys and mode.proc_rank == 0:
    trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
    trace_command = (f'nsys profile --output {trace_output} ...')
elif enable_rocprof and mode.proc_rank == 0:
    trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
    trace_command = (f'rocprofv2 --hip-trace --kernel-trace --plugin json -d {trace_output} -- ')

# torch.distributed / mpi — same selection logic repeated again below
```

**Issue:** The nsys/rocprof selection-and-quoting logic is copy-pasted almost verbatim three times. This duplication pre-existed for nsys alone; the PR doubles its footprint by adding an equivalent rocprofv2 branch in each of the three copies instead of factoring out a shared helper.

**Impact:** Any future change to trace-flag formatting, quoting behavior, or a third profiler tool requires editing (and keeping in sync) three near-identical blocks, increasing the risk of drift/bugs where one branch is fixed but the other two are missed.

**Recommendation:** Extract a private helper, e.g. `self.__build_trace_command(trace_dir, rocprof_trace_dir, benchmark_name, suffix, enable_nsys, enable_rocprof)`, returning the formatted prefix/command string, and call it from all three branches.

superbench/runner/runner.py
```diff
@@ -135,12 +136,25 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+        enable_rocprof = os.environ.get('SB_ENABLE_ROCPROF', '') == '1'
+        rocprof_trace_dir = os.environ.get('SB_ROCPROF_TRACE_DIR', self._sb_output_dir)
+
         mode_command = exec_command
         if mode.name == 'local':
-            trace_command = (
-                f'nsys profile --output {trace_dir}/{benchmark_name}_{mode.proc_rank}_traces '
-                f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
-            ) if enable_nsys and mode.proc_rank == 0 else ''
+            trace_command = ''
```
polarG, line 192: **[NON-BLOCKING] Inconsistent naming for the same concept: `trace_command` vs `trace_prefix`**

**Evidence:**
```python
# local / mpi
trace_command = ''
...
# torch.distributed
trace_prefix = ''
```

**Issue:** `local` and `mpi` name the variable `trace_command` while the functionally-equivalent `torch.distributed` branch uses `trace_prefix` for the same role.

**Impact:** Minor readability cost; makes the triplicated logic harder to visually diff across branches.

**Recommendation:** Standardize on one name across all three branches, ideally as part of extracting a shared helper.

superbench/runner/runner.py
```diff
@@ -159,23 +173,41 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+                    f'rocprofv2 --hip-trace --kernel-trace --plugin json '
+                    f'-d {trace_output} -- '
+                )
 
             mode_command = (
-                f'{nsys_prefix}'
+                f'{trace_prefix}'
                 f'torchrun'
                 f' --no_python --nproc_per_node={mode.proc_num} {torch_dist_params}{exec_command}'
                 f' superbench.benchmarks.{benchmark_name}.parameters.distributed_impl=ddp'
```
polarG, line 225: **[NON-BLOCKING] `benchmark_name` remains unquoted elsewhere in the same shell command**

**Evidence:**
```python
mode_command = (
    f'{trace_prefix}'
    f'torchrun'
    f' --no_python --nproc_per_node={mode.proc_num} {torch_dist_params}{exec_command}'
    f' superbench.benchmarks.{benchmark_name}.parameters.distributed_impl=ddp'
    f' superbench.benchmarks.{benchmark_name}.parameters.distributed_backend=nccl'
)
```

**Issue:** `benchmark_name` is quoted when building the new rocprof/nsys trace path but left unquoted where it's spliced directly elsewhere in the same shell command. Pre-existing pattern, not introduced by this PR, but the PR's own quoting effort makes the gap more visible.

**Impact:** Low risk in typical single-operator deployments since `benchmark_name` comes from the user's own `sb.config.yaml`, but is a latent gap if configs are ever shared/templated across users.

**Recommendation:** Not required to block this PR; consider consistent quoting/validation for `benchmark_name` in a follow-up.

superbench/runner/runner.py
```diff
@@ -135,12 +136,25 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
         enable_nsys = os.environ.get('SB_ENABLE_NSYS', '') == '1'
         trace_dir = os.environ.get('SB_NSYS_TRACE_DIR', self._sb_output_dir)
 
+        # Enable rocprofv2 profiling based on environment variable
+        enable_rocprof = os.environ.get('SB_ENABLE_ROCPROF', '') == '1'
+        rocprof_trace_dir = os.environ.get('SB_ROCPROF_TRACE_DIR', self._sb_output_dir)
```
polarG, line 180: **[NON-BLOCKING] No validation of `SB_ROCPROF_TRACE_DIR`/`SB_NSYS_TRACE_DIR` path content**

**Evidence:**
```python
enable_rocprof = os.environ.get('SB_ENABLE_ROCPROF', '') == '1'
rocprof_trace_dir = os.environ.get('SB_ROCPROF_TRACE_DIR', self._sb_output_dir)
```

**Issue:** The trace directory is taken directly from an environment variable with no path validation (e.g. rejecting `..` traversal or requiring an allowed root) before being used to build `-d`/`--output` targets.

**Impact:** Independent of the shell-injection issue noted separately, a malicious/misconfigured value could redirect trace output outside the intended directory, overwriting arbitrary writable files.

**Recommendation:** Optionally normalize/validate the resulting path (`Path(...).resolve()` + confirm it stays under an expected root). Low priority given this is operator-supplied configuration.

superbench/benchmarks/model_benchmarks/pytorch_base.py
polarG: **[NON-BLOCKING] Nested boolean/`getattr` condition reduces readability**

**Evidence:**
```python
if not (torch.cuda.is_available() and (
    getattr(torch.version, 'cuda', None) is not None or getattr(torch.version, 'hip', None) is not None
)):
```

**Issue:** The double-negative (`not (... and (... or ...))`) combined with two inline `getattr` calls is harder to scan than the previous single-condition check.

**Impact:** Minor cognitive overhead for future maintainers of this GPU-vendor detection logic.

**Recommendation:** Extract to a small named helper, e.g. `self._is_supported_gpu_vendor()` returning a bool.

polarG (changes requested):

superbench/benchmarks/model_benchmarks/pytorch_base.py
shcho: Verified on MI300x (ROCm 6.3.4, PyTorch ROCm build) during trace generation testing — ProfilerActivity.CUDA and event.cuda_time are aliases for HIP in PyTorch ROCm builds, and the generated profiler JSON contains valid GPU timings. I'll add a short comment noting the alias so future readers don't hit the same concern. CI coverage on ROCm hardware would be ideal but is out of scope for this PR.

superbench/runner/runner.py
shcho: rocprofv2 with --plugin json does emit per-PID output files (one per process it observes), and torchrun-spawned workers appeared as distinct PIDs in the output. No code changes.

superbench/runner/runner.py
shcho: Previous PR for NSYS also has no tests. Should address both in the future PRs.

superbench/benchmarks/model_benchmarks/pytorch_base.py
shcho: Previous PR for NSYS also has no tests. Should address both in the future PRs.

superbench/runner/runner.py
shcho: non-blocking.

superbench/runner/runner.py
shcho: Non blocking.

superbench/runner/runner.py
shcho: Non blocking.

superbench/benchmarks/model_benchmarks/pytorch_base.py
shcho: Non blocking.

superbench/runner/runner.py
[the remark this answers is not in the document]
shcho: Addressed in commits.

superbench/runner/runner.py
```diff
@@ -159,23 +173,41 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
-            ) if enable_nsys else ''
+            trace_prefix = ''
+            if enable_nsys:
+                trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_traces')
+                trace_prefix = (
+                    f'nsys profile --output {trace_output} '
+                    f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
+                )
+            elif enable_rocprof:
+                trace_output = shlex.quote(f'{rocprof_trace_dir}/{benchmark_name}_traces')
```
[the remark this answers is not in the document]
shcho: Addressed in commits.

superbench/runner/runner.py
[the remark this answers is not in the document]
shcho: Addressed in commits.

superbench/runner/runner.py
```diff
@@ -135,12 +136,25 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
+
         mode_command = exec_command
         if mode.name == 'local':
-            trace_command = (
-                f'nsys profile --output {trace_dir}/{benchmark_name}_{mode.proc_rank}_traces '
-                f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
-            ) if enable_nsys and mode.proc_rank == 0 else ''
+            trace_command = ''
+            if enable_nsys and mode.proc_rank == 0:
+                trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
```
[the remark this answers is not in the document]
shcho: Addressed in commits.

superbench/runner/runner.py
shcho: Addressed in commits.

superbench/runner/runner.py
shcho: Addressed in commits.

superbench/runner/runner.py
shcho: Addressed in commits.

superbench/runner/runner.py
```diff
@@ -159,23 +173,41 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
...
                 f' superbench.benchmarks.{benchmark_name}.parameters.distributed_backend=nccl'
             )
         elif mode.name == 'mpi':
-            trace_command = (
-                f'nsys profile --output {trace_dir}/{benchmark_name}_{mode.proc_rank}_traces '
-                f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
-            ) if enable_nsys else ''
+            trace_command = ''
+            if enable_nsys:
+                trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_{mode.proc_rank}_traces')
```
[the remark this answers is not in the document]
shcho: Addressed in commits.

superbench/runner/runner.py
```diff
@@ -159,23 +173,41 @@ def __get_mode_command(self, benchmark_name, mode, timeout=None):
                 '--nnodes=$NNODES --node_rank=$NODE_RANK --master_addr=$MASTER_ADDR --master_port=$MASTER_PORT '
             )
 
-            nsys_prefix = (
-                f'nsys profile --output {trace_dir}/{benchmark_name}_traces '
-                f'--backtrace none --sample none --force-overwrite true --cpuctxsw none --trace cuda,nvtx '
-            ) if enable_nsys else ''
+            trace_prefix = ''
+            if enable_nsys:
+                trace_output = shlex.quote(f'{trace_dir}/{benchmark_name}_traces')
```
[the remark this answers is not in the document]
shcho: Addressed in commits.

superbench/runner/runner.py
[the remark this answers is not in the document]
gusui-msft: I think the earlier Copilot suggestion may be incorrect for ROCm 6.3.4: `rocprofv2` does not support a bare `--` and treats it as an unknown option.

Could you verify this with the ROCm 6.3.4 `rocprofv2` in the target image? The expected syntax is:

```text
rocprofv2 <options> <executable> [args...]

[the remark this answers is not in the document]
Comment (shcho) [reply superbench/runner/runner.py]: Thanks for the comment. The bare -- is removed in the latest commit.

**Review by gusui-msft — APPROVED** [review]

File: superbench/runner/runner.py [hunk 22]
```diff
@@ -25,6 +26,17 @@
...
 
+def _quote_for_bash_lc(value):
+    r"""Quote a value so it is safe both for the shell and for embedding inside an outer bash -lc '...' string.
+
+    ``shlex.quote`` wraps values containing whitespace or shell metacharacters in single quotes. When the
+    resulting command is later interpolated into ``bash -lc '{command}'``, those single quotes would
+    terminate the outer single-quoted context. Escape any single quotes as ``'\''`` so the value survives
+    both quoting layers.
+    """
+    return shlex.quote(value).replace("'", "'\\''")
```
polarG, line 37: **[NON-BLOCKING] Add regression coverage for nested-shell quoting**

This helper is the fix for the prior nested-shell quoting defect, but the existing runner tests never enable either profiler and therefore do not exercise it. Please add focused cases for spaces, apostrophes, command-substitution text, and semicolons, covering both the Docker `bash -lc` and skip-Docker `bash -c` command shapes.

polarG (approved):

superbench/runner/runner.py
shcho: Addressed in the latest commit.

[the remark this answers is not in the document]
shcho: Addressed in the latest commit.

tests/runner/test_runner.py
```diff
@@ -270,6 +272,69 @@ def test_run_empty_benchmarks(self):
...
+        and skip-Docker (``bash -c``) dispatch paths.
+        """
+        cases = [
+            '/tmp/traces',
+            '/tmp/my traces/foo_traces',
+            "/tmp/mia's dir/foo_traces",
+            '/tmp/$(id)/foo_traces',
+            '/tmp/`whoami`/foo_traces',
+            '/tmp/foo;rm -rf x/foo_traces',
+            '/tmp/foo|bar/foo_traces',
```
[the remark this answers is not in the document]
shcho: Addressed in the latest commit.

tests/runner/test_runner.py
```diff
@@ -270,6 +272,69 @@ def test_run_empty_benchmarks(self):
...
+        ]
+        # The two wrapper shapes actually used by run_sys_info / _run_proc.
+        wrapper_shapes = [
+            "bash -lc 'echo {token}'",
+            "bash -c 'echo {token}'",
+        ]
```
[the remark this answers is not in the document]
shcho: Addressed in the latest commit.

PR #837, opened 2026-07-29
Dockerfile - Add ROCm6.4 dockerfile

polarG:
**Description**
Adds a ROCm 6.4.4 SuperBench Docker image targeting AMD MI300X (gfx942), based on #812, plus fixes discovered while building and verifying the image end-to-end on MI300X hardware.

**Major Revision**
- Add `dockerfile/rocm6.4.x.dockerfile` — new ROCm 6.4.4 image (Ubuntu 24.04, Python 3.12, PyTorch 2.7.1, gfx942), base `rocm/pytorch:rocm6.4.4_ubuntu24.04_py3.12_pytorch_release_2.7.1`.
- OFED upgraded to DOCA-Host 25.10 user-space (`doca-ofed-userspace=3.2.3-019000`), matching the target host's OFED version; MLNX_OFED standalone tarballs are gone upstream.
- Embedded Docker pinned to client-only `29.6.2` (extracts only the `docker` client binary, dropping dockerd/containerd/runc/shims to shrink the CVE surface — SuperBench only shells out to the Docker client, never runs its own daemon).
- Add `flex` to the apt package list (fixes an OpenMPI "developer build"/`autogen.pl` failure that requires it).
- Set `AMDGPU_TARGETS=gfx942` explicitly (previously unset, a regression vs `rocm6.3.x.dockerfile`) — without a GPU present at `docker build` time, hipcc silently defaulted to `gfx906`, which caused incorrect zero-copy reads / `gpu-copy-bw:correctness` failures on MI300X.
- Bump `third_party/rccl-tests` submodule to `0039629` (adds a missing `<cstring>` include required for ROCm 6.4's HIP headers, ROCm/rccl-tests#132).
- Fix `third_party/Makefile` rccl-tests build: repeat the HIP platform defines (`-x hip -D__HIP_PLATFORM_AMD__ -D__HIPCC__`) inside the `HIPCUFLAGS` override, since setting `AMDGPU_TARGETS` causes the override to replace (not append to) rccl-tests' own `HIPCUFLAGS`, which otherwise drops those defines and breaks the HIP headers.
- Register `rocm6.4` in `.github/workflows/build-image.yml`.

**Minor Revision**
- CMake `if/else` shell fix (missing `;` before `else`).
- `rm -rf .git` cleanup in the final `make postinstall` step (image size).
- `apt-get clean` in the apt/OFED install blocks.

**Verification**
A full `sb run` was executed on an 8x MI300X host: 460/462 executor runs passed. The 3 failures (`resnet` rendezvous port collision, `gpu-copy-bw:perf` OOM, `gpu-copy-bw:correctness`) were root-caused; the correctness failure was the `AMDGPU_TARGETS` issue above, confirmed in-container with the full correctness suite passing 321/321 after rebuilding with `gfx942`. The other two failures are fixed outside this Dockerfile (runner/config changes) and are not blockers for this PR.


setup.py
```diff
@@ -19,8 +19,10 @@
 print(f'Python {sys.version_info.major}.{sys.version_info.minor} detected.')
 if sys.version_info[:2] < (3, 11):
     import pkg_resources
+    # setuptools 66+ no longer supports Python 3.7, newer runtimes work with any recent version.
```
gusui-msft: setuptools 66+ still supports Python 3.7, and this `setup.py` change is unrelated to this PR anyway, consider moving it out.

dockerfile/rocm6.4.x.dockerfile
```diff
@@ -0,0 +1,245 @@
...
+    apt-get clean && \
+    rm -rf /tmp/*
+
+ARG NUM_MAKE_JOBS=64
+
+# Target GPU architecture(s) for ROCm micro-benchmark builds (space-separated).
+# Without this, AMDGPU_TARGETS is empty and hipcc defaults to gfx906 at build time
+# (no GPU is present during `docker build`), producing wrong-arch kernels that run
+# incorrectly on MI300X — e.g. gpu-copy-bw:correctness fails its CheckBuf data check.
+# Override at build time with: --build-arg AMDGPU_TARGETS="gfx90a gfx942 gfx950".
```
WenqingLan1: The Dockerfile documents AMDGPU_TARGETS as a build-time override and gives gfx90a gfx942 gfx950 as an example. The value is passed to hipBLASLt's -a option, but HIPBLASLT_LOGIC_FILTER remains fixed to aquavanjaram/gfx942/**/*. TransformerEngine is also always built with NVTE_ROCM_ARCH=gfx942. A caller selecting gfx90a or gfx950 can therefore receive an image whose hipBLASLt logic is filtered to a different architecture and whose TransformerEngine extension is compiled for gfx942.

Suggested fix: Derive the hipBLASLt logic selection and NVTE_ROCM_ARCH from the requested targets, or remove the advertised general override and validate that AMDGPU_TARGETS is exactly gfx942. If TransformerEngine supports only one target per build, fail early when multiple targets are supplied and document that constraint. Add a cheap build-configuration check that covers the default and at least one rejected or supported non-default value.

setup.py
polarG: Removed the unrelated setuptools change and restored the original setup.py requirement in 84a953a6.

dockerfile/rocm6.4.x.dockerfile
polarG: Updated the build configuration in 84a953a6. AMDGPU_TARGETS is normalized and propagated to hipBLASLt and TransformerEngine; single-target hipBLASLt logic filters are selected for gfx908, gfx90a, and gfx942, while mixed targets use the full requested architecture list. AOTriton remains enabled for gfx942 and is disabled for lower or mixed targets. Added a Docker/GPU-independent configuration test covering the default, lower, mixed, and irregular-whitespace inputs.

gusui-msft (approved):

WenqingLan1 (approved):

PR #838, opened 2026-07-29
Bugfix - Use volatile accesses for HIP zero-copy loads/stores in gpu-copy-bw

polarG:
**Description**

The `gpu-copy-bw` micro-benchmark's SM-copy kernels use two helper functions, `FetchULong2()` and `StoreULong2()` (`superbench/benchmarks/micro_benchmarks/gpu_copy_performance/gpu_copy.cu`), to read/write 16-byte chunks from/to zero-copy host-mapped (or peer) memory. On the CUDA path these already use `ld.volatile.global` / `st.volatile.global` inline PTX — uncached, non-reorderable accesses, which is the correct pattern for memory that can be concurrently written by another device/host and isn't guaranteed to be cache-coherent through the normal load/store path. On the HIP path, the same functions instead did a **plain, cacheable** pointer dereference (`v.x = p->x; v.y = p->y;`), so the compiler was free to cache, reorder, or otherwise not treat these as true "read what's actually there right now" accesses.

This PR reinterprets the pointer as `volatile ulong2*` before dereferencing on the HIP path, forcing an uncached, non-reorderable access that matches the CUDA behavior.

**Investigation context**

This fix was originally written to explain a `gpu-copy-bw:correctness` failure (`CheckBuf: Memory check failed` on `cpu_to_gpu0_by_sm_under_numa0`) observed during a full SuperBench run on an MI300X/ROCm 6.4.4 host. Follow-up in-container investigation (rebuilding `gpu_copy` from source with an explicit `--offload-arch=gfx942`) showed that **the actual failure in that run was caused by a missing `AMDGPU_TARGETS` build flag** in the ROCm 6.4 Dockerfile, which made `hipcc` silently default to `gfx906` when no GPU was present at `docker build` time — the resulting binary then ran wrong-architecture kernels on the gfx942 MI300X hardware and produced bad reads. That Dockerfile fix is tracked separately in #837.

With the correct `gfx942` arch, the correctness suite passes 321/321 with or without this `volatile` change, so it was **not** the root cause of the specific failure observed. It is included here anyway as a genuine correctness hardening: the HIP path for zero-copy memory access should not silently diverge from the CUDA path's use of volatile/uncached semantics, and leaving it as a plain access is latent-bug-prone on any HIP toolchain/compiler version that becomes more aggressive about caching such reads.

**Major Revision**
- `FetchULong2()` (HIP branch): read through a `const volatile ulong2 *` instead of a plain `const ulong2 *`.
- `StoreULong2()` (HIP branch): write through a `volatile ulong2 *` instead of a plain `ulong2 *`.

**Minor Revision**
- None — single-file, four-line-of-intent change (10 insertions / 4 deletions after re-indentation).

**Testing**
- Rebuilt `gpu_copy` in-container on an 8x MI300X host with `--offload-arch=gfx942`; correctness suite (`--check_data`) passes 321/321 both with and without this change once the Dockerfile's `AMDGPU_TARGETS` fix (#837) is applied.
- No behavior change expected on CUDA (this only touches the `#if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)` branch).


gusui-msft (approved):

setup.py
```diff
@@ -19,8 +19,10 @@
 print(f'Python {sys.version_info.major}.{sys.version_info.minor} detected.')
 if sys.version_info[:2] < (3, 11):
     import pkg_resources
+    # setuptools 66+ no longer supports Python 3.7, newer runtimes work with any recent version.
```
gusui-msft: setuptools 66+ still supports Python 3.7, and this setup.py change is unrelated to this PR anyway, consider moving it out.

superbench/benchmarks/micro_benchmarks/gpu_copy_performance/gpu_copy.cu
```diff
@@ -599,8 +599,11 @@ int DestroyEvent(BenchArgs *args) {
 // https://github.com/ROCmSoftwarePlatform/rccl/blob/5c8380ff5b5925cae4bce00b1879a5f930226e8d/src/collectives/device/common_kernel.h#L268
 inline __device__ void FetchULong2(ulong2 &v, const ulong2 *p) {
 #if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)
-    v.x = p->x;
-    v.y = p->y;
+    // Use a volatile access so the compiler does not cache/reorder this zero-copy read of
+    // host-mapped (or peer) memory, matching the uncached ld.volatile.global on the CUDA path.
+    const volatile ulong2 *vp = reinterpret_cast<const volatile ulong2 *>(p);
```
gusui-msft: `ulong2` is `HIP_vector_type<unsigned long, 2>`, whose `x`/`y` live in the `HIP_vector_base` base class, and clang's HIP device codegen drops `volatile` on member access through that base — so this hunk generates the same `gfx942` code as before. Worth confirming on your 6.4.4 toolchain before rewording the comment.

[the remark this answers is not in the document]
polarG: Addressed in b9514bfb and refined in 5d76241a. The comments now describe compiler elision/reordering only and no longer claim cache-bypass or uncached semantics. ROCm 6.4.4 gfx942 device IR retains two volatile 64-bit loads.

setup.py
polarG: Addressed in b9514bfb. The unrelated setup.py change was removed; the PR now changes only gpu_copy.cu.

superbench/benchmarks/micro_benchmarks/gpu_copy_performance/gpu_copy.cu
polarG: Addressed in b9514bfb and 5d76241a. The HIP path now takes separate volatile scalar member pointers instead of accessing members through volatile ulong2 pointers. With ROCm 6.4.4 clang 19 at -O2 for gfx942, the revised helpers emit two volatile 64-bit loads and two volatile 64-bit stores; repeated same-address probes also retain both accesses.

superbench/benchmarks/micro_benchmarks/gpu_copy_performance/gpu_copy.cu
```diff
@@ -614,8 +617,11 @@ inline __device__ void FetchULong2(ulong2 &v, const ulong2 *p) {
 // https://github.com/ROCmSoftwarePlatform/rccl/blob/5c8380ff5b5925cae4bce00b1879a5f930226e8d/src/collectives/device/common_kernel.h#L276
 inline __device__ void StoreULong2(ulong2 *p, ulong2 &v) {
 #if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)
-    p->x = v.x;
-    p->y = v.y;
+    // Use a volatile access so the compiler does not cache/reorder this zero-copy write to
+    // host-mapped (or peer) memory, matching the uncached st.volatile.global on the CUDA path.
+    volatile ulong2 *vp = reinterpret_cast<volatile ulong2 *>(p);
```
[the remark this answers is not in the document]
polarG: Addressed in b9514bfb and refined in 5d76241a. The store comment no longer claims an uncached write; it states only that scalar volatile accesses prevent compiler elision or reordering. ROCm 6.4.4 gfx942 device IR retains two volatile 64-bit stores.

superbench/benchmarks/micro_benchmarks/gpu_copy_performance/gpu_copy.cu
```diff
@@ -599,8 +599,12 @@ int DestroyEvent(BenchArgs *args) {
...
 inline __device__ void FetchULong2(ulong2 &v, const ulong2 *p) {
 #if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)
-    v.x = p->x;
-    v.y = p->y;
+    // HIP drops volatile on ulong2 member access, so access its scalar storage directly
+    // to prevent the compiler from eliding or reordering this zero-copy read.
+    const volatile unsigned long *x_pointer = &p->x;
+    const volatile unsigned long *y_pointer = &p->y;
+    v.x = *x_pointer;
+    v.y = *y_pointer;
```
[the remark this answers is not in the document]
polarG: Addressed in 5d76241a. The scalar pointer types are now derived with unparenthesized decltype(p->x) and decltype(p->y) instead of hardcoding unsigned long. This preserves explicit volatile qualification without adding type-traits dependencies. ROCm 6.4.4 gfx942 at -O2 emits the same two volatile loads and stores, and compile-time checks confirm the unparenthesized decltype yields each member's declared scalar type.

PR #844, opened 2026-08-12
Pin GitHub Actions to full-length commit SHAs

danfiedler-msft:
## Summary

This PR pins GitHub Actions to full-length commit SHAs for improved security and reproducibility and adds a 7 day cooldown to Dependabot configuration for GitHub Actions. This work is described in more detail at https://aka.ms/action-pinning.

## Why?

Pinning actions to commit SHAs prevents supply-chain attacks where a tag could be moved to point to malicious code. This is a recommended security best practice per the [GitHub Actions security hardening guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

This change mitigates the risk of tag retargeting to malicious code as seen in incidents like the [tj-actions/changed-files action compromise](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised) or [codfish/semantic-release-action compromise](https://www.stepsecurity.io/blog/supply-chain-compromise-codfish-semantic-release-action) and improves the integrity and reproducibility of the CI/CD pipeline.

## What changed?

**Action pinning:** Third-party action references in `.github/workflows/` that used mutable tag-based references (e.g., `actions/checkout@v4`) have been updated to full-length commit SHAs with a version comment (e.g., `actions/checkout@<sha> # v4`) using the [pinact](https://github.com/suzuki-shunsuke/pinact) tool. References that were already pinned to a SHA, or that used immutable release tags, were left unchanged.

**Dependabot configuration:** `.github/dependabot.yml` has been updated to ensure a `github-actions` package-ecosystem section is present with a `cooldown` configuration (`default-days: 7`). This groups Dependabot PRs for GitHub Actions and enforces a minimum 7-day cooldown between updates. If the file did not exist, it was created. If a `github-actions` section already existed, only the `cooldown` block was added or its `default-days` value was increased to 7 if it was lower. The 7-day cooldown provides a window for the community to detect and report compromised releases before they are automatically proposed as updates, reducing exposure to supply-chain attacks via newly published malicious versions.

## Is this safe to merge?

Yes. The pinned SHAs correspond to the same commits that the existing tags pointed to. No behavioral changes are introduced. You can verify the pinned SHA value using the GitHub REST API (e.g., the commit hash for `actions/checkout@v7` can be found in the `sha` property in the JSON response for `GET https://api.github.com/repos/actions/checkout/commits/v7`).

## Additional Information

For more information, please see https://aka.ms/action-pinning

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -28,15 +28,15 @@ jobs:
           - javascript
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@a37ce9120846195fa4ece8f58b268e6043cb2f26 # v3.7.0
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@c4dd10e44af883a891fe31ced449bcb4a6728b9b # v3.37.6
         with:
           languages: ${{ matrix.language }}
       - name: Autobuild
-        uses: github/codeql-action/autobuild@v3
+        uses: github/codeql-action/autobuild@c4dd10e44af883a891fe31ced449bcb4a6728b9b # v3.37.6
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@c4dd10e44af883a891fe31ced449bcb4a6728b9b # v3.37.6
   analyze-cpp:
     name: CodeQL analyze cpp
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -58,10 +58,10 @@ jobs:
         with:
```
polarG, line 58: ### [SHOULD-FIX] Pin the remaining mutable CMake branch

**Location:** [.github/workflows/codeql-analysis.yml](.github/workflows/codeql-analysis.yml#L57) -- `jobs.analyze-cpp` / `Setup CMake`

**Evidence:**
```yaml
      - name: Setup CMake
        uses: lukka/get-cmake@latest
        with:
          cmakeVersion: '3.20.0'
```

**Issue:**
Verified: `latest` is the upstream branch `refs/heads/latest`, not an immutable release tag. At review time it resolves to `fffaaafeea488556c2c12dad60690008bc1caacb`, also the peeled commit of tag `v4.4.2`.

This reference predates the PR, but leaving it unpinned omits an in-scope action from the pinning migration. Dependabot's cooldown governs proposed version updates; it does not delay changes to a branch resolved when a workflow runs.

**Impact:**
The C++ CodeQL job can execute changed upstream action code without a repository commit or reviewed dependency update. A compromised upstream branch could tamper with analysis in a job granting `security-events: write`. This is a risk scenario, not an observed compromise.

**Recommendation:**
Pin the action to the verified equivalent full commit SHA, preserving the existing CMake input:
```yaml
      - name: Setup CMake
        uses: lukka/get-cmake@fffaaafeea488556c2c12dad60690008bc1caacb # v4.4.2
        with:
          cmakeVersion: '3.20.0'
```

PR #852, opened 2026-09-10
Benchmarks: Fix cuDNN storage/compute types and FP16 initialization

gusui-msft:
**Description**
Fix the cuDNN configuration's convolution compute type without overwriting the tensor storage type. Previously, requesting FP16 storage with FP32 compute changed `input_type_` to FP32, so dispatch could execute FP32 while the configuration still described FP16.

Correct FP16 dispatch also exposes the half initializer's reversed host-to-device copy arguments. This PR includes the necessary initialization fix.

**Major Revision**
- Make `set_conv_type()` and `get_conv_type()` use `conv_type_` independently of `input_type_`.
- Seed FP16 initialization with the requested seed and copy the initialized host buffer to the device buffer in the correct direction.

**Minor Revision**
- Keep the patch runtime-only: two files, +4/-3; no regression-test files or CMake changes.
- Leave benchmark cases, algorithms, timing and metric names unchanged. Corrected FP16 results require new baselines; this is not a latency-variance fix.

**Validation**
- The unchanged runtime fixes were previously validated for independent storage/compute types and seeded FP16 initialization on CUDA 13.0 / cuDNN 9.12.
- Rebased onto `main` at `5411041d809c7573984acdc2a989057b589b8366` on September 11; published head is `bb7391a9e6796604f0e250e1ef38d84595b31e17`.
- Verified that the complete two-file patch and runtime file contents are identical to the pre-rebase revision, CMake matches current main, the regression-test file is absent, and `git diff --check` passes.
- No new GPU run was performed for the test removal or rebase. No automated regression test is included.


gusui-msft:
CI triage for `bb7391a9e6796604f0e250e1ef38d84595b31e17` (2026-09-13, observed 05:28 UTC):

- **The required CUDA check remains unresolved.** The public [main branch protection metadata](https://api.github.com/repos/microsoft/superbenchmark/branches/main) lists `cuda-unit-test` as required. [Build 59534](https://msrasrg.visualstudio.com/3ca1f950-a27b-4701-86fd-35f043130c82/_build/results?buildId=59534) failed, but its public annotation contains only `Bash exited with code '1'.` at build-log line 595. An authorized `msrasrg` pipeline reader is needed to identify the failed task and provide a non-sensitive error/traceback excerpt. This is not yet classified as either a cuDNN regression or an infrastructure failure. The existing cuDNN integration test does exercise the defaults, FP16 custom configuration and auto-algorithm path.
- **ROCm 6.4 failed before the changed source was introduced.** The original job failed in `make ... -C third_party rocm ...` with exit 2. In the unchanged [Dockerfile at this head](https://github.com/microsoft/superbenchmark/blob/bb7391a9e6796604f0e250e1ef38d84595b31e17/dockerfile/rocm6.4.x.dockerfile), that layer follows `ADD third_party third_party` and precedes `ADD . .` and SuperBench compilation. The nested dependency error is not exposed by the annotation, so no specific OOM/network diagnosis is claimed. The owner's [attempt 2](https://github.com/microsoft/superbenchmark/actions/runs/34625817961/attempts/2) is running on the same head; no duplicate rerun was submitted.
- **Ansible is still queued.** [Build 59535](https://msrasrg.visualstudio.com/3ca1f950-a27b-4701-86fd-35f043130c82/_build/results?buildId=59535) has no public queue explanation. Its pipeline uses pool `SuperBench CI` with demand `ansible-agent`; pool/authorization state needs a pipeline owner. Neither Ansible nor ROCm 6.4 is in the currently exposed required-check list. The GitHub image rerun does not restart either Azure pipeline.

Requested review from the repository's `microsoft/superbench-dev` code-owner team. The patch remains runtime-only, two files (+4/-3), with no source or workflow changes during triage. This comment does not establish merge readiness: CUDA must be diagnosed and the required checks and any independent approval requirements satisfied first.

